PR #127, opened 2013-10-20
Use GLOB_SUBST with $arg ($~arg) and check suffix aliases

### LINKED ISSUE #126: error/warning "_zsh_highlight_main_highlighter:36: bad assignment" when completing "!$", where the expansion starts with a "="
After

    % vi =foo.sh

tab-completion of `!$` results in an error:

    % !$_zsh_highlight_main_highlighter:36: bad assignment
    =foo.sh

Using `set -x` it is caused by this:

    +_zsh_highlight_main_highlighter:38> alias '=foo.sh'
    _zsh_highlight_main_highlighter:38: bad assignment

The argument is considered to be an alias from this:

    res=$(LC_ALL=C builtin type -w $arg 2>/dev/null)

    % builtin type -w =foo.sh 
    sh: alias

    % builtin type =foo.sh   
    sh is a suffix alias for zsh-mime-handler

The first issue is, that `=foo.sh` should get resolved first (absolute path of "foo.sh", if any), and then looked at by `type`.

The second minor issue appears to be that the suffix aliases setup by zsh-mime-setup should not get considered to be an alias, if they are a command/executable file maybe.

    autoload -U zsh-mime-setup
    zsh-mime-setup

A fix for the second problem might be to use "-p" with `type`.

blueyed:
$~arg resolves "=foo" and suffix aliases might be used in general or via
zsh-mime-setup.

Fixes #126

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -93,11 +93,21 @@ _zsh_highlight_main_highlighter()
      if [[ -n ${(M)ZSH_HIGHLIGHT_TOKENS_PRECOMMANDS:#"$arg"} ]]; then
       style=$ZSH_HIGHLIGHT_STYLES[precommand]
      else
-      res=$(LC_ALL=C builtin type -w $arg 2>/dev/null)
+      # ~: use GLOB_SUBST to resolve "=foo.sh" (https://github.com/zsh-users/zsh-syntax-highlighting/issues/126)
+      if [[ $arg[1] == '=' ]]; then
+        local arg_resolved=$~arg
+      else
+        local arg_resolved=$arg
+      fi
+      res=$(LC_ALL=C builtin type -w $arg_resolved 2>/dev/null)
       case $res in
         *': reserved')  style=$ZSH_HIGHLIGHT_STYLES[reserved-word];;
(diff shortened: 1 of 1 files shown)


danielshahaf:
I don't think we can just use `$~arg` because it may run arbitrary code, e.g.,

```
% s='*.txt(e*REPLY=foo*)' 
% print -l ${(z)s}        
*.txt(e*REPLY=foo*)
% print $~s               
foo foo foo foo
```

I don't understand why `nonomatch` is needed.

blueyed:
Re `nonomatch` I wrote the following in 2013, but cannot recall why and when (I thought it was needed):

> Otherwise it may fail hard if e.g. "=1" is not found while editing the prompt. (https://github.com/blueyed/zsh-syntax-highlighting/commit/b4df380e267ecac56e6ae9ea3098a15039361f6d)

That might be fixed in another way by now?!

danielshahaf:
Yes, I saw that log message but didn't understand from it what the problem was. So let's put the `nonomatch` patch on hold until someone can argue for it.

The first patch in this series can't be merged as-is (due to use of `$~arg` as mentioned above, and since the parent issue has consensus that the right fix is to make `type -w` smarter), so let's close this PR, and once the `type -w` improvements lands in zsh, we can fix zsh-syntax-highlighting to make use of that.

PR #136, opened 2014-01-31
highlight path separators

blaenk:
This adds support for highlighting path separators. Let me know what you think.


danielshahaf:
(Sorry for the late response; I was only granted push access here last week.)

Quick review: the implementation seems reasonable.

- Using `:=` for `$style_override` seems wrong, since that sets the variable _after_ the place that tests its value.
- In `mkdir -p /etc/foo/bar/baz`, path separators are not highlighted if there are two or more non-existent path components. From a user's point of view, this would be nice to have; from an implementor's, I'm not sure how complicated this would be to implement (without too many false positives, complexity, etc).

I currently focus on bugfixes, not on new features.

danielshahaf:
This was resubmitted as #260 and merged.  Thanks for the patch!

PR #149, opened 2014-09-25
README: make installation step 1 more accurate

lathan:
You cannot download just the "script". You need the entire directory/repo.

This is the same #145 but with fixed link.


danielshahaf:
I changed the patch a little to avoid a URL in the middle of a paragraph. Thanks for the patch, @lathan.

Issue #30, opened 2011-01-05
Rewrite in C


sorin-ionescu:
This script is extremely slow. Once we are satisfied with the level of highlighting, it will most likely have to be rewritten in C as a ZSH patch and enabled via a shell option such as `setopt syntax_highlighting`. Hopefully, it will be accepted.

zachliu:
It IS slow:
```
num  calls                time                       self            name
-----------------------------------------------------------------------------------
 1)    1          34.27    34.27   50.39%     34.27    34.27   50.39%  _zsh_highlight_bind_widgets
 2)    1           9.95     9.95   14.63%      9.95     9.95   14.63%  compinit
 3)    1           9.04     9.04   13.29%      8.21     8.21   12.07%  _zsh_highlight_load_highlighters
 4)    8           4.58     0.57    6.74%      4.58     0.57    6.74%  compdef
 5)    1           3.32     3.32    4.89%      3.32     3.32    4.89%  include
 6)    5           2.74     0.55    4.04%      2.74     0.55    4.04%  add-zsh-hook
 7)    1           1.09     1.09    1.60%      1.09     1.09    1.60%  is-at-least
 8)    4           3.47     0.87    5.11%      1.01     0.25    1.49%  complete
 9)    3           0.70     0.23    1.04%      0.70     0.23    1.04%  path_ladd
10)    1           0.62     0.62    0.91%      0.62     0.62    0.91%  va
```

Issue #153, opened 2014-10-07
Unusably slow on Cygwin


strugee:
I have zsh-syntax-highlighting enabled via Antigen, and on Cygwin, it's unusably slow. I think this might be a recent regression, but I'm not sure.

```
% zsh --version
zsh 5.0.6 (x86_64-unknown-cygwin)
```

I'd provide more information, but I'm not sure where to begin debugging this. Please let me know what I can do, and I'll try to add some information.

strugee:
@ferenczy what OS are you on? I think we're talking about separate issues. On Cygwin, this package is slow for me even when _I'm_ typing.

Also, some more information: it appears that this may be triggering a bug in something else. When I SSH into my Debian Jessie server, which has this enabled, everything works just fine. It's only on a local shell that it's slow. Maybe it has something to do with Cygwin's zsh build?

Ardakilic:
+1 , I'm using [babun](http://babun.github.io/) on Windows 8.1 pro x64, it comes with zsh 5.06, and after enabling it's quite slow for me, too. Pasting 200 character takes more than 5 seconds.

strugee:
I can no longer reproduce this.

    % zsh --version
    zsh 5.0.8 (x86_64-unknown-cygwin)
    % cd ~/.antigen/repos/*zsh-syntax-highlighting*
    % git rev-parse HEAD
    3dc574190071ef59054f6d693132410b37aa3ae4

How about other people? If nobody pipes up, I'll close this issue.

strugee:
I should mention that I recently freshly installed Cygwin on a fresh Windows box.

tylercal:
Still terribly slow for me:
```
> zsh --version
zsh 5.0.8 (x86_64-unknown-cygwin)
```

Used a downloaded version with latest commit a8afbdf2f548b6782d2a33d40c6d72464e23beee and also tried the release from Sept 2014 @strugee mentioned (3dc574190071ef59054f6d693132410b37aa3ae4)

Kleptine:
Horribly slow on a fresh installation of Windows 10 + Babun. Everything installed today. 

Typing just feels awful. Anyone have a workaround?

kiyeon:
It's the same phenomenon.
Why is this?

```sh
$ zsh --version
zsh 5.1.1 (i686-pc-cygwin)
```

inet-pwnZ:
I expirenced the same problems but if i go into bash and then back to zsh it gets faster

[...]

PR #155, opened 2014-12-13
Proposal of text pasting performance issue solution

ferenczy:
As I described in [this comment](https://github.com/zsh-users/zsh-syntax-highlighting/issues/30#issuecomment-66862941), pasting of text to the command line with enabled highlighter is quite pain, and even more on Cygwin. Originally, pasting of 31 characters (what's pretty short) to the command line **took 5 seconds. With my change it's instant.** But it's rather proposal than real solution. I'm not shell programming ninja, so I would welcome any advices to make it better.

**It works like this**: since a condition `[[ $PENDING -gt 0 ]] && return $ret` is not working (at least not on Cygwin), highlighting is executed after each character is inserted while pasting. Therefore I have added another condition after the previous one, which measures delta time between previous and current keystroke. If it's less than 200 ms, highlighting does not continue.

Period of 200 ms was empirically verified like ideal. Shorter periods didn't work.

**There is one thing, which is not ideal in my proposal.** If user presses 2 last keystrokes faster than in 200 ms, highlighting is not executed after the last keystroke. He for example presses few characters and then TAB, which completes the command, in less than 200 ms and ends up with non highlighted command. He has to press for example a space after that to highlight the previous command.

It would be fixed by for example executing of `_zsh_highlight` function periodically, even if no key was pressed (e.g. once per second). Or maybe you'll have some better idea.

In any case, my current solution proposal is way better for me than before. I would rather press a space sometimes to highlight current line than have to wait crazy time to finish a paste. But as I wrote, it's not ideal.

It's interesting that `test-perfs.zsh` measures the same times for both cases, with and without my solution. But pasting is definitely instant.

Also one note to a difference between Cygwin and Linux in relation to performance. Windows doesn't have a `fork`. Creating of a new process is very expensive on Windows. Therefore shell scripts which call a lot of another binaries (or scripts) are much more slower on Cygwin. I have used `date` to get epoch in my first attempt and it was even worse than before. So I have switched to *datetime* Zsh module and using  of `$EPOCHREALTIME` variable instead of `date` command. Condition is evaluated instantly now.

I have created very simple script:
````
COUNTER=50
while [ $COUNTER -gt 0 ]; do
    echo `date +%s`
    let COUNTER=COUNTER-1
done
````

...and there are times it took to execute it under both Cygwin and Linux:

<pre>
system         time
----------------------------
Cygwin         6.091 sec
Linux (VM)     0.062 sec
-----------------------------
</pre>

According to this, it runs 100 times faster on Linux (even in VM) than on Cygwin. But scripts without external calls take similar time on both systems. It's just about creating of a process.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -33,6 +33,9 @@
 # Core highlighting update system
 # -------------------------------------------------------------------------------------------------
 
+# Load datetime Zsh module for measuring time delta between keystrokes
+zmodload zsh/datetime
+
 # Array declaring active highlighters names.
 typeset -ga ZSH_HIGHLIGHT_HIGHLIGHTERS
 
@@ -53,6 +56,11 @@ _zsh_highlight()
 
   # Do not highlight if there are pending inputs (copy/paste).
   [[ $PENDING -gt 0 ]] && return $ret
+
+  # Do not highlight if the time delta from previous keystroke is less than 200 ms
+  [[ -z "$LAST_INVOKE_EPOCH" ]] && typeset -g LAST_INVOKE_EPOCH=$EPOCHREALTIME
+  [[ $(( $EPOCHREALTIME - $LAST_INVOKE_EPOCH )) -lt 0.2 ]] && return $ret
+  typeset -g LAST_INVOKE_EPOCH=$EPOCHREALTIME
 
   # Reset region highlight to build it from scratch
   region_highlight=();


danielshahaf:
It would be interested to know whether slowness is still observed under zsh-5.1: that release added support for bracketed pasting, which change the way the line editor processes pasted characters.

ferenczy:
OK, I'll try it as soon as zsh 5.1 is released for Cygwin (there is version 5.0.8 currently).

danielshahaf:
At the zsh level, `$PENDING` is available on all systems that support the `FIONREAD` ioctl.  Can you check on your system, whether the preprocessor conditional block in zsh's `Src/Zle/utils.c:noquery()` is even compiled?  Perhaps the problem is that zsh on cygwin needs to do something to make that ioctl visible (include another header, for example)?

The relevant callsite is `zle_params.c:get_pending()`.

danielshahaf:
Closing per https://github.com/zsh-users/zsh-syntax-highlighting/issues/153#issuecomment-222057200

PR #159, opened 2015-01-27
Multiline

danielshahaf:
The 'main' highlighter does not highlight certain multiline constructs
correctly.  By code inspection, it disregards $PREBUFFER entirely, so any
multiline construct where the start of a line is not a new command would be
misparsed and mishighlighted.  This patch series fixes a few such cases.
Regression tests are included.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -102,7 +102,7 @@ _zsh_highlight_main_highlighter()
     local substr_color=0
     local style_override=""
     [[ $start_pos -eq 0 && $arg = 'noglob' ]] && highlight_glob=false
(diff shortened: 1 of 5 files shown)


danielshahaf:
I've pushed another commit to the branch that handles the before-the-newline part of multiline strings.  (I accidentally left that part out during a rebase.)

danielshahaf:
The fourth patch 384be15b130a54d19e3d0183b645edd440645ccb does not fix the case of a literal newline immediately followed by a space. It does fix the case of a literal newline immediately followed by non-whitespace. I presume the whitespace-skipping logic of `$start_pos` is the issue.

danielshahaf:
> The fourth patch 384be15 does not fix the case of a literal newline immediately followed by a space.

Fixed by the just-pushed commit c2b9327b0763f3457fb09db17e22ee9e1e024792.

PR #151, opened 2014-09-25
zle: don't override *delete* and *kill*

eras:
This restores the ability to consecutively delete (ie.) words and
have them come back consecutively in the next yank.

PR #160, opened 2015-02-05
zle: don't override kill-word and backward-kill-word

### LINKED ISSUE #150: Consecutive delete-words do not get yanked consecutively
Steps to reproduce:

1) enter "hello hello" to prompt
2) press ^W ^W, both hellos are removed
3) press ^Y and witness one 'hello' coming back

Expected behavior: "hello hello" would come back (works if zsh-syntax-highlighting is not enabled)

I sort of hoped #143 would have fixed this but no :).

vincentbernat:
Otherwise, we can't do `echo hello hello^W^W^Y` and get back `echo hello
hello` (we get `echo hello` instead because ZLE doesn't detect two
consecutive calls to backward-kill-word).

I don't add more to this commit because other `*-kill-*` are less likely
to be used twice in a row when working with the shell. But maybe,
`kill-line` could be added too.

Closes #150.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -139,7 +139,7 @@ _zsh_highlight_bind_widgets()
 
   # Override ZLE widgets to make them invoke _zsh_highlight.
   local cur_widget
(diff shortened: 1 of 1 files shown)


vincentbernat:
In fact, this is a duplicate of #151 that I didn't see...

nicoulaj:

PR #163, opened 2015-03-07
highlight comments when interactive_comments is on

sonnym:
A fairly straightfoward little change that highlights comments if the `interactive_comments` option is enabled.  Please let me know if you would like to see any additional changes to this pull request before merging it.  Thanks for a great project!

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -52,6 +52,7 @@
 : ${ZSH_HIGHLIGHT_STYLES[double-quoted-argument]:=fg=yellow}
 : ${ZSH_HIGHLIGHT_STYLES[dollar-double-quoted-argument]:=fg=cyan}
 : ${ZSH_HIGHLIGHT_STYLES[back-double-quoted-argument]:=fg=cyan}
+: ${ZSH_HIGHLIGHT_STYLES[comment]:=fg=yellow}
 : ${ZSH_HIGHLIGHT_STYLES[assign]:=none}
 
 # Whether the highlighter should be called or not.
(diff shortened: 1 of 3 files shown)


danielshahaf:
(Sorry for the late response; I was only granted push access here last week.)

I've had a quick look; review points so far:

1. Your branch has diverged with master; could you please rebase?  (My apologies for those conflicts; they are mostly with my changes, it seems. Let me know if you have questions about the upstream changes.)

2. Could you add a test for `echo "foo # bar" baz`, testing that `baz` is _not_ highlighted as a comment?

3. False positive: `echo "foo # bar` currently highlights `# bar` as a comment, but shouldn't. (This may change once you rebase past ff4d402e9a587e0e8f8400a3aeac40a8f73215e5, which was merged to master in ea81e3df6ed14dd5dcb72c1685cf83fd5b0e47a9.)

4. False positive: in

        % alias foo\#bar=echo\ hello
        % foo#bar

    both `#` are parsed as comment starts, but neither of them is.  I realize getting all cases right might be hard, but could the situation be improved?  At least the first — a backslash-escaped comment character — ought to be reasonably common.

5. What about multiline buffers?  For example, if you type `echo # foo<^V><^J>bar`, the `bar` should be highlighted as a command or unknown command, not as a comment.  (If you use `bindkey -e`, read `<^V><^J>` as `M-Enter`.)  You can test that, too, by setting `$PREBUFFER` in the test.

6. The `${(z)…}` splitter currently (zsh-5.1) behaves as follows:

        % setopt interactivecomments
        % print -rl - ${(q-)${(z):-"foo#bar"}}
        'foo#bar'

   Shouldn't it parse `foo` and `#bar` as separate tokens, at least when the option is set?  (If yes, that would be a bug in zsh itself, not in zsh-syntax-highlighting.)

7. Please mention the new `$ZSH_HIGHLIGHT_STYLES` key in README where all recognized keys are listed.

8. You should declare `local comment_start`.

Thanks for the patch!  I look forward to merging it.

danielshahaf:
Fixed.  I used the tests from your patch, but I removed `comment-embedded.zsh` because it was incorrect (the parser considers `foo#bar` as a single word even in INTERACTIVE_COMMENTS mode).  Thanks again!

PR #170, opened 2015-05-19
Fix syntax error with zsh 5.0.7-dev-3

danielshahaf:
Commit 35168 (52aeb9aaeb4799b760138a7c34b18ede4b47242a) in zsh changed the parsing of case patterns, resulting in parse errors.  Change the syntax to be compatible with zsh both before and after that commit.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -228,7 +228,7 @@ _zsh_highlight_main_highlighter_highlight_string()
             fi
               (( varflag = 0 )) # End of variable
             ;;
-      ([^a-zA-Z0-9_]))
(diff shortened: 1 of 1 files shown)


blueyed:
:+1: 

I can confirm that this is an issue and that the fix works.

blueyed:
But the zsh version is 5.0.8-dev-3.

danielshahaf:
Nope, 5.0.7-dev-3 is correct.  (The foo-dev-N versions come after foo, not before it.)

danielshahaf:
The parsing incompatibility was fixed upstream by commit 35248 (zsh-users/zsh@afb78f5d142786169817709e6ec4c48637bcae93), so this patch is no longer needed.  Closing.

Issue #165, opened 2015-04-13
vi-mode: visual mode: no background for commands


blueyed:
When using vi-keybindings and entering visual mode (`v`), the visual selection gets a background color / standout mode, but this does not work with some parts when using zsh-syntax-highlighting, e.g. commands and invalid command:

In the below screenshot, the whole line is visually selected, but it is not visible for `ls` and `lss`:

![selection_065](https://cloud.githubusercontent.com/assets/9766/7115262/33783f2e-e1e7-11e4-8760-34016619efaa.png)

rosshadden:
:+1:

danielshahaf:
In your example, printing `$region_highlight` from a widget shows:

```
region_highlight=( '0 2 fg=green' '3 6 underline' '7 10 none' '11 12 none' '13 16 fg=red,bold' )
```

Combining that with your screenshot shows reverse video only applies to the portions that have an `fg=…` attribute.

I can reproduce this without zsh-syntax-highlighting:

```
$ zsh -f
% bindkey -v
% f() { region_highlight=('0 2 fg=green' '2 4 underline') }
% zle -N f
% bindkey -a ^T f
% abcdef<Esc>v0<^T>    # cdef in reverse video, ab in green on normal-background-color
```

I reckon zsh itself should show the `ab` in reverse video too.  @blueyed @rosshadden Could you report this to the zsh project?

blueyed:
> Should zsh turn the fg=green to bg=green?

Seems sensible in the case of reverse?!

danielshahaf:
> Seems sensible in the case of reverse?!

It seems both you and I expect `$region_highlight` to be applied first, and the MARK..CURSOR region's highlighting to be applied on top of that. (Yes, the names are confusing.) However, it would not be unreasonable of zle to do things the other way around: first to apply the MARK..CURSOR region's highlighting and then to apply `$region_highlight` on top of that, so widgets can override the MARK..CURSOR region's highlighting.

The MARK..CURSOR's region highlighting, which defaults to `standout` (i.e., reverse video), is specified by `$zle_highlight`.

The man page is silent on the interaction between `$zle_highlight` and `$region_highlight`. I think that should be clarified by upstream. Once it is clarified whose responsibility it is to reverse the video of the `ls` and `lss` (in your example), that party — either zle or zsh-syntax-highlighting — can implement a fix.

Makes sense?

danielshahaf:
[...]

PR #175, opened 2015-07-23
Bracketed paste highlight

danielshahaf:
zsh has grown a feature whereby pasted text can be highlighted. These patches are required for the new upstream feature to work when zsh-syntax-highlighting has been loaded.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -139,7 +139,7 @@ _zsh_highlight_bind_widgets()
 
   # Override ZLE widgets to make them invoke _zsh_highlight.
   local cur_widget
-  for cur_widget in ${${(f)"$(builtin zle -la)"}:#(.*|_*|orig-*|run-help|which-command|beep|yank*)}; do
+  for cur_widget in ${${(f)"$(builtin zle -la)"}:#(.*|_*|orig-*|run-help|bracketed-paste|which-command|beep|yank*)}; do
     case $widgets[$cur_widget] in
 
       # Already rebound event: do nothing.
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -139,7 +139,7 @@ _zsh_highlight_bind_widgets()
 
   # Override ZLE widgets to make them invoke _zsh_highlight.
   local cur_widget
-  for cur_widget in ${${(f)"$(builtin zle -la)"}:#(.*|_*|orig-*|run-help|bracketed-paste|which-command|beep|yank*)}; do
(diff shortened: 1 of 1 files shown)


danielshahaf:
zsh has grown a feature whereby pasted text can be highlighted. These patches are required for the new upstream feature to work when zsh-syntax-highlighting has been loaded.

danielshahaf:
That's not quite right: it causes highlighting not to be updated until the first keypress after the paste. This can happen in two cases:

- When `$zle_highlight` _does_ specify paste highlighting, and `RBUFFER` is not empty (i.e., the paste is not at the end of the command-line), and the paste changes the parsing of `$RBUFFER`; for example, if the command-line is `echo` and a `"` is pasted before the first letter, the letters `echo` should become yellow (part of a double-quoted string).

- When `$zle_highlight` does _not_ specify highlighting for pastes, the pasted text itself should be highlighted. For example, pasting `echo foo` into an empty command-line should highlight the `echo` in green underline (without an additional keypress).

danielshahaf:
This `$zle_highlight`/`$region_highlight` interaction might be related to [the disussion on issue #165](https://github.com/zsh-users/zsh-syntax-highlighting/issues/165#issuecomment-136598904).

danielshahaf:
The proposed fix of #165 at https://github.com/zsh-users/zsh-syntax-highlighting/issues/165#issuecomment-138689186 



The fix at (https://github.com/danielshahaf/zsh-syntax-highlighting/commit/0e31d6e1a298adfce9596b7048f207c9940d495d) also fixes this issue, but it requires  (which will be included in zsh-5.1.1 and newer).

danielshahaf:
The [patch for issue #165](https://github.com/danielshahaf/zsh-syntax-highlighting/commit/0e31d6e1a298adfce9596b7048f207c9940d495d) (linked [at that issue](https://github.com/zsh-users/zsh-syntax-highlighting/issues/165#issuecomment-138689186)) also fixes this issue, if REGION_ACTIVE,CURSOR,MARK are replaced by YANK_ACTIVE,YANK_START,YANK_END respectively. The last three are zle parameters added to zsh in https://github.com/zsh-users/zsh/commit/c088906423ba22eef8d57c84b396f8d5be83ca45, which is due to be released as part of zsh-5.1.1.

danielshahaf:
Proposed patch: https://github.com/danielshahaf/zsh-syntax-highlighting/commit/e5c2a88db023600d7aeacbb2b7d640e0c8671a83

(The branch containing that patch also contains the aforementioned proposed fix to #165.)

danielshahaf:
> Proposed patch: danielshahaf@e5c2a88

Merged in https://github.com/zsh-users/zsh-syntax-highlighting/commit/51102bf83fdfdca34f5be82eb3cac5518be64305.

PR #177, opened 2015-08-23
Added quick and dirty install Makefile

ktgeek:
for systems where the package manager doesn't have an existing package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,15 @@
+NAME=zsh-syntax-highlighting
+
+DIRS=highlighters tests
+INSTALL_DIRS=`find $(DIRS) -type d 2>/dev/null`
+DOC_FILES=*.md
+ZSH_FILES=*.zsh
+
+PREFIX?=/usr/local
+
+SHARE_DIR=$(DESTDIR)$(PREFIX)/share/$(NAME)
+
+install:
+	for dir in $(INSTALL_DIRS); do mkdir -p $(SHARE_DIR)/$$dir; done
+	find $(DIRS) -type f -print0 | xargs -0 -I % cp % $(SHARE_DIR)/%
+	cp -r $(DOC_FILES) $(ZSH_FILES) $(SHARE_DIR)


danielshahaf:
Three points:
1. The `-r` flag to `cp` on line 15 is not needed.
2. The variable name `$(DIRS)` is so generic that I wonder whether some make implementation out there might be using it as a reserved variable name with associated special semantics (like `$path`, lowercase, is special under `zsh`). So perhaps use a different name?  That will also help avoid confusion between `$(DIRS)` and `$(INSTALL_DIRS)`.
3. How about adding an `@echo Add "'source $(PREFIX)/share/$(NAME)/zsh-syntax-highlighting.zsh'"` message to the end?

Other than that, looks good to me. Thanks for the patch!

danielshahaf:
Another nice-to-have would be to have the installed tree identify the installed version: which tag or git revision the installed tree is.

ktgeek:
I like it.  I haven't had a chance to revisit it yet, but I should have time soon

-- 
 Keith T. Garner                                       kgarner@kgarner.com
                "Make no little plans; they have no magic to
                   stir men's blood." - Daniel H. Burnham

phy1729:
Perhaps a smaller makefile can do such as https://github.com/phy1729/zsh-syntax-highlighting/blob/4acb8f3c8c66910e8c8ddc78bcee4be3fe49da1f/Makefile ?

danielshahaf:
@phy1729 I like your simpler implementation, but the default `make` on Debian doesn't define the `$(INSTALL)` variable.  Adding `INSTALL?=install -c` fixes the issue for me:

```
remotes/phy1729/makefile 4b46ff01d76d 
% make -n
d /usr/local/share/zsh-syntax-highlighting
cp -r zsh-syntax-highlighting.zsh highlighters /usr/local/share/zsh-syntax-highlighting
                                                                                                                                                         
remotes/phy1729/makefile 4b46ff01d76d 
% make -n INSTALL='install -c'
install -c -d /usr/local/share/zsh-syntax-highlighting
cp -r zsh-syntax-highlighting.zsh highlighters /usr/local/share/zsh-syntax-highlighting
```

danielshahaf:
@phy1729 About the `test` target [[1]](https://github.com/phy1729/zsh-syntax-highlighting/blob/4b46ff01d76d191d2e986845bd8e1a5b8e5e70b7/Makefile): it's a welcome addition; I suggest to have `make test` exit non-zero if any test failed (that is, if any invocation of `test-highlighting.zsh` returned non-zero).

danielshahaf:
I've merged [@phy1729's branch](https://github.com/phy1729/zsh-syntax-highlighting/tree/makefile) in 75d62b965711d5987d4acb221d0be95c9b4662b0.

@ktgeek, thanks for the patch; I've opted to go with @phy1729's version since it was a simpler implementation.  Thanks again!

ktgeek:
No worries! I'm glad a good implementation is there.  Sorry I never got a chance to go back and clean up mine.

-- 
 Keith T. Garner                                       kgarner@kgarner.com
                "Make no little plans; they have no magic to
                   stir men's blood." - Daniel H. Burnham

danielshahaf:
> I've merged @phy1729's branch in 75d62b9.

Issue #218 was filed to add documentation for the new makefile targets.

PR #214, opened 2015-10-19
README.md: install with oh-my-zsh in a single command

vitorgalvao:
No need to `cd` to the directory (especially if you’re installing this plugin as part of a bigger script).

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,7 @@ How to install
 
 * Download the script or clone this repository in [oh-my-zsh](http://github.com/robbyrussell/oh-my-zsh) plugins directory:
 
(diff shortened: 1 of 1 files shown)


danielshahaf:
Agreed, but you missed a path component in the last argument to `git clone`, didn't you?

```
+        git clone git://github.com/zsh-users/zsh-syntax-highlighting.git ~/.oh-my-zsh/custom/plugins/zsh-syntax-highlighting
```

vitorgalvao:
Nice catch. It is now corrected.

danielshahaf:
Thanks for the patch!

Issue #148, opened 2014-09-10
Highlighting completion state


mhanne:
I'm not quite sure how this is supposed to work, so I'm probably missing some ingenious feature here..

But to me the highlighting on the first word of the command seems a bit random. For example if I type `mplayer`, it stays yellow/underlined until I type the `r`, then turns green. When I type `pidgin`, it turns red after typing the `g`, and then green when the whole word is there. What's that supposed to tell me?

My suggestion would be to leave the word white by default, change it to red if it doesn't match anything in the path (i.e. pressing `<tab>` gives you nothing), and green if it's a valid command. Now that we have the yellow color freed up, we can use that to indicate when there is a unique completion already matching our command (telling me I can simply press `<tab>` now), and maybe even something else to "warn" me that I typed more than would strictly be necessary for a unique completion)

For example, `pidg<tab>` completes to pidgin, so `pid` should be white (since it has other possible completions like pidof), `pidg` yellow (now I can simply press `<tab>`), `pidgi` magenta (or whatever, to tell me I wouldn't have had to type that last character) and `pidgin` green.

danielshahaf:
> But to me the highlighting on the first word of the command seems a bit random. For example if I type mplayer, it stays yellow/underlined until I type the r, then turns green. When I type pidgin, it turns red after typing the g, and then green when the whole word is there. What's that supposed to tell me?

I guess your cwd is `~` when you try this. When you type `mplaye` (sic), that helpfully gets matched against your `~/.mplayer` directory and highlighted as yellow+underline by the `path_approx` style. Pidgin isn't affected because it uses `~/.purple`.

The yellow underline is `path_approx` highlighting, which has no right to kick in for the command word, unless the "approximately spelled file" is a executable, or is a readable directory and `AUTO_CD` is set. So that's a bug.

In any case, I like your suggestion of revisiting the command-word highlighting. What possible states need the highlighting distinguish?

How about [valid command / invalid command] × [unique completion / ambiguous completion]? All four combinations (valid/unique, valid/ambiguous, invalid/unique, invalid/ambiguous) are possible. "Ambiguous" would mean that there is more than one possible completion, including the word-so-far itself, that is a valid command word (e.g., `vi` is ambiguous since either `vi` or `vim` could be intended). If the command word is valid and the cursor isn't inside the command word, then the "unambiguous" variant would be used. There is a fifth option, "invalid command / no completions", which would be red/bold like today.

[...]

PR #244, opened 2015-11-27
[#148] command being typed as neither green nor red

danielshahaf:
Related to #148: this patch adds a highlight style for the command word, while it is being typed (similar to `path_prefix`).

The style used is yellow underline, the same style the now-removed `path_approx` used.

The intention is to avoid the unknown-token style whilst inputting the command. Future changes may fine tune this further (as discussed in #148), for example, only use the new style when the input-so-far is a prefix of a command or function name.

### DIFF
--- a/docs/highlighters/main.md
+++ b/docs/highlighters/main.md
@@ -26,6 +26,7 @@ This highlighter defines the following styles:
 * `precommand` - precommand modifiers (e.g., `noglob`, `builtin`)
 * `commandseparator` - command separation tokens (`;`, `&&`)
 * `hashed-command` - hashed commands
+* `command-being-typed` - prefix of a command/function/alias that isn't one itself (e.g., `sourc`, `ifconfi`)
 * `path` - existing filenames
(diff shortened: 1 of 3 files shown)


danielshahaf:
Perhaps this should use `$#LBUFFER` (= `$CURSOR`) instead of `$#BUFFER`.

nicoulaj:

phy1729:
Can this use `command_prefix` or `arg0_prefix` to be consistent with other in progress styles?

vegerot:
I still think a `command_prefix` or `arg0_prefix` would be very useful to me and others.  Is there any progress on allowing this?

danielshahaf:
Thanks for the ping. Milestoning: small change, but destabilizing, so 0.8.0. (0.7.0 is closed for destabilizing changes.)

danielshahaf:
Cross-referencing #695.

This is really a sub-issue of #695, so taking this out of the milestone.  #695 is still milestoned (not sure whether it'd be fixed before the release, but it's on the shortlist).

Issue #65, opened 2011-08-02
Sourcing multiple times fails with zsh 4.3.11


simonheimlicher:
I really like this script and have been using it for several months now.

With the new version that is separated into multiple files, sourcing the script multiple times (through sourcing ~/.zshrc after changes) causes problems. 

Running a plain zsh ("zsh -fl"), then sourcing the main script twice via
```source ~/.zsh/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
source ~/.zsh/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh

``````
causes the following issues:

- zsh 4.3.11 on Mac OS X Lion: Segmentation Fault
- zsh 4.3.11 on Debian stable: ```_zsh_highlight_widget_accept-line: job table full or recursion limit exceeded```

Backtrace from GDB on Mac OS X Lion:
```book% source ~/.zsh/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
Program received signal EXC_BAD_ACCESS, Could not access memory.
Reason: KERN_PROTECTION_FAILURE at address: 0x00007fff5f3fffe0
0x000000010001242a in execode ()
(gdb) bt
#0  0x000000010001242a in execode ()
#1  0x0000000100015e20 in execlist ()
#2  0x000000010001623b in execlist ()
#3  0x000000010001590b in execlist ()
#4  0x000000010001237d in execode ()
#5  0x0000000100016a88 in runshfunc ()
#6  0x0000000100016ea8 in doshfunc ()
#7  0x0000000100308ce8 in execzlefunc ()
#8  0x00000001003146b6 in bin_zle ()
#9  0x000000010000d57f in execbuiltin ()
#10 0x0000000100014a1d in execode ()
#11 0x0000000100015e20 in execlist ()
#12 0x000000010001623b in execlist ()
#13 0x000000010001590b in execlist ()
#14 0x000000010001237d in execode ()
#15 0x0000000100016a88 in runshfunc ()
#16 0x0000000100016ea8 in doshfunc ()
#17 0x0000000100308ce8 in execzlefunc ()
#18 0x00000001003146b6 in bin_zle ()
#19 0x000000010000d57f in execbuiltin ()
#20 0x0000000100014a1d in execode ()
#21 0x0000000100015e20 in execlist ()
#22 0x000000010001623b in execlist ()
#23 0x000000010001590b in execlist ()
#24 0x000000010001237d in execode ()
[...]

PR #255, opened 2015-12-31
properly rebind more relevant zle widgets

m0vie:
- Just because a widget starts with _ does not mean it should be
  skipped when rebinding. The only reason widgets need to be skipped
  is when their function name start with _zsh_highlight_widget_*,
  which is checked later.

  Example:
  _expand_alias (^Xa) needs to be wrapped.

- Skipping when $cur_widgets == user:$cur_widget is wrong.
  This skips widgets whose function name is identical to
  the widget name and is not related to _zsh_highlight_widget_.

  Example:
  expand-absolute-path is a widget whose function name is also
  expand-absolute-path. No reason why this should be ignored.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -200,7 +200,7 @@ _zsh_highlight_bind_widgets()
 
   # Override ZLE widgets to make them invoke _zsh_highlight.
   local cur_widget
-  for cur_widget in ${${(f)"$(builtin zle -la)"}:#(.*|_*|orig-*|run-help|which-command|beep|set-local-history|yank)}; do
+  for cur_widget in ${${(f)"$(builtin zle -la)"}:#(.*|orig-*|run-help|which-command|beep|set-local-history|yank)}; do
     case $widgets[$cur_widget] in
 
       # Already rebound event: do nothing.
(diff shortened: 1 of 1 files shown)


danielshahaf:
> Skipping when $cur_widgets == user:$cur_widget is wrong.

Agreed.  The code was added in 776453cb5b6922a61c01fb51f1c0bdbaf73b79d3 with comment `# Already rebound event`, but it also matches widgets such as as `edit-command-line`. The change looks correct, and it makes `edit-command-line` update highlighting.

> Just because a widget starts with _ does not mean it should be skipped when rebinding.

This was added in 186d80054a40262b9db967aeaf483a28027de19f, which was a fix for #65.  I assume the `_*` pattern can be removed since the case pattern suffices...  Will have to test this (and check that #65 isn't reintroduced).

danielshahaf:
Perhaps this should be split to two commits?  I know it's a trivial patch, but splitting it will make future bisects (should they be needed) easier.

danielshahaf:
Thanks for splitting the patch — I hadn't seen that until now (github didn't notify me of the forced push).

danielshahaf:
Merged.  Apologies for the delay, it's entirely my fault, the patch was flawless.

(GitHub think the patch branch is "conflicting" because I fixed a log message typo before merging.)

Thanks for the patch!

PR #256, opened 2015-12-31
respect deliberately empty entries in zle_highlight

m0vie:
Don't force the default value when the user deliberately sets an
entry in zle_highlight to the empty string.

For example

  zle_highligh(paste:)

to disable paste highlighting completely.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -113,33 +113,13 @@ _zsh_highlight()
     done
 
     # Re-apply zle_highlight settings
-    () {
-      if (( REGION_ACTIVE )) ; then
-        # zle_highlight[region] defaults to 'standout' if unspecified
-        local region="${${zle_highlight[(r)region:*]#region:}:-standout}"
-        integer start end
-        if (( MARK > CURSOR )) ; then
-          start=$CURSOR end=$MARK
-        else
-          start=$MARK end=$CURSOR
-        fi
-        region_highlight+=("$start $end $region")
-      fi
-    }
-    # YANK_ACTIVE is only available in zsh-5.1.1 and newer
-    (( $+YANK_ACTIVE )) && () {
-      if (( YANK_ACTIVE )) ; then
-        # zle_highlight[paste] defaults to 'standout' if unspecified
-        local paste="${${zle_highlight[(r)paste:*]#paste:}:-standout}"
-        integer start end
-        if (( YANK_END > YANK_START )) ; then
-          start=$YANK_START end=$YANK_END
-        else
-          start=$YANK_END end=$YANK_START
-        fi
-        region_highlight+=("$start $end $paste")
-      fi
-    }
+
+    # region
+    (( REGION_ACTIVE )) && _zsh_highlight_apply_zle_highlight region standout "$MARK" "$CURSOR"
+
+    # yank / paste (zsh-5.1.1 and newer)
+    (( $+YANK_ACTIVE )) && (( YANK_ACTIVE )) && _zsh_highlight_apply_zle_highlight paste standout "$YANK_START" "$YANK_END"
+
 
     return $ret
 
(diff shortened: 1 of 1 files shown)


danielshahaf:
Using an empty string after the colon appears to be undocumented, but it's used in the wild, so we might as well support it.

Please avoid declaring the same variable `local` twice, it spams stdout:

```zsh
% () { local x; local x } 
x=''
```

As a matter of style, I'd have stripped the `region:` prefix earlier and made the condition use `!= ""` (or just `-n`), I think "string is not empty" is easier to read than "string is not foo".

danielshahaf:
By the way, preexisting problem: `region_highlight+=("$start $end $region")` should use `${(q)region}` instead of `$region`.

m0vie:
> Please avoid declaring the same variable local twice, it spams stdout:

Done.

> As a matter of style, I'd have stripped the region: prefix earlier and made the condition use != "" (or just -n), I think "string is not empty" is easier to read than "string is not foo".

I did that initially, however then there would be no way to distinguish between the case where it was deliberately set to the empty string, or where the token does not exist at all inside zle_highlight:

`zle_highlight=(foo:bar)` would result in the empty string and `zle_highlight=(paste:)` would, too. But these cases need to be handled differently: In the former, since `paste:` was not given at all, we use the default value, in the latter, we do nothing. I couldn't think of a shorter way to test those cases, so I went with the current solution.

m0vie:
> By the way, preexisting problem: `region_highlight+=("$start $end $region")` should use `${(q)region}` instead of `$region`.

Why is that? What could `$region` possibly contain that would need to be quoted?
And what about `_zsh_highlight_main_add_region_highlight`? Should be be calling this instead?

danielshahaf:
> I did that initially, however then there would be no way to distinguish between the case where it was deliberately set to the empty string, or where the token does not exist at all inside zle_highlight:

Right, so the code has to distinguish three cases: the `(r)` didn't match (`[[ $region == "" ]]`); the `(r)` matched a `"paste:"` with nothing after the colon; the `(r)` matched a `"paste:foo"`.

> [...] I couldn't think of a shorter way to test those cases, so I went with the current solution.

Yes, it's short.  And it's correct.  However, I don't find it very clear.  I'd rather figure out some way to rewrite the condition to make it easier to read.  (Or, failing that, a big fat comment explaining what's going on.)

Perhaps...

```zsh
local region
if region="${zle_highlight[(r)region:*]}";
   [[ -z $region ]] || { region=${region#"region:"}; [[ -n $region ]] }; then
```

danielshahaf:
> Why is that? What could `$region` possibly contain that would need to be quoted?

Sorry, that was a mistake on my part, I think.  I did recall the `comment*.zsh` tests needed a (q-) guard (as in: the test was incorrect without that), but that was in $expected_region_highlight [which is parsed using ${(z)...}), not in $region_highlight (which doesn't; ref. set_region_highlight() in the zsh C source).

> And what about `_zsh_highlight_main_add_region_highlight`? Should be be calling this instead?

No.  That function expects PREBUFFER-relative arguments, but MARK and CURSOR are relative to BUFFER, not to PREBUFFER.

(Also, the driver shouldn't be calling into individual highlighters, but if that were the only problem, moving the helper function to the driver would have addressed it.)

danielshahaf:
I haven't seen 1411dc0ac73ff6aa83eb03ae54bb4b3742b2fdb8 until now.  @m0vie, @github does not notify me when you force-push a PR'd branch unless you also comment on the PR at the same time.  Is this a problem in my account settings?  If not, please make sure to add comments, otherwise I won't see your updates timely.  In any case, sorry for the delay, let me review it now.

1. The $+YANK_ACTIVE check is needed to distinguish zsh<5.1.1 from zsh≥5.1.1.  I know that older zsh's will evaluate (( YANK_ACTIVE )) to false, but I don't want to rely on that, for several reasons: style/clarity, forward compatibility (what if zsh-5.3 makes that variable default to `1`?), and robustness (what if somebody patches the code to do something when YANK_ACTIVE is zero? If there's no explicit settedness check, the patched code might unintentionally run for zsh<5.1.1).  So I'd rather keep that check.

2. Please write a docstring for `_zsh_highlight_apply_zle_highlight_on_top`, and mention that the first positional argument mustn't contain pattern metacharacters (since line 138 interprets $entry as a pattern, and you can't write ${(b)entry} for backwards compatibility reasons).

3. [Line 148](https://github.com/m0vie/zsh-syntax-highlighting/commit/1411dc0ac73ff6aa83eb03ae54bb4b3742b2fdb8#diff-4ed73021a0fa2ff4e78a9909f4fc2f9bR148): please enforce precedence by adding {} around the disjunction.  I am not sure whether it's correct as written, but I'd like to have {} anyway (as in `{ foo || bar } && baz`) for code readability.

Thanks for the revised patch, and sorry again for not noticing it earlier.

danielshahaf:
Once this is fixed, we can fix #267 as well.  (No semantic dependency, but they touch the same code, so serializing the fixes would avoid needless conflicts.)

m0vie:
Updated. Squashed the commits and fixed the three issues.

danielshahaf:
Merged, thanks!

Issue #245, opened 2015-11-30
redrawhook issues


danielshahaf:
The following issues should be fixed by the zsh redrawhook branch (https://github.com/zsh-users/zsh-syntax-highlighting/issues/154#issuecomment-144541575):
#90 #150 #154 #183 #203 #377

(confirmed for #90 and #150; the rest are known to be widget-wrapping-related issues, and the branch does away with widget wrapping)

~~The following issue probably won't be:~~
#40

_edit: Upstream committed a patch that, in in conjunction with this issue, fixes #40, q.v._

The following issues might be fixed by using the `zle -f yank` / `zle -f kill` feature already in upstream zsh:
#150 #183

_edit_: #375 is also reported to be fixed by redrawhook branch, and #500 is suspected to be (not tested).

danielshahaf:
The zsh redrawhook branch was merged upstream in https://github.com/zsh-users/zsh/commit/c3ea3ffa730037714fd1bb23b6fc1174b379368f, to appear in zsh-5.2.1.

danielshahaf:
Note potential issue with `print -zr`: [workers/38499](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=38499)

danielshahaf:
Upstream is considering introducing an `add-zsh-hook`-like API for hooking to `zle-line-pre-redraw`: [workers/38499](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=38569)

danielshahaf:
> Upstream is considering introducing an add-zsh-hook-like API for hooking to zle-line-pre-redraw: workers/38499

Done: [workers/38670](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=38670)

danielshahaf:
I have drafted a patch that uses workers/38670 here: https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:mikachu-redrawhook/v3  At the moment it requires zsh master + two patches from workers@, but I expect both of these patches will be applied to zsh master soon enough.

See also #261 and https://github.com/zsh-users/zsh-syntax-highlighting/pull/261#issuecomment-168234362 for two implementations that precede add-zle-hook-widget.

danielshahaf:
Status update:

There are two candidate implementations:

[...]

PR #257, opened 2015-12-31
Do not highlight in isearch widgets

m0vie:
As long as zle_highlight for the isearch region can't be applied
properly *after* zsh-syntax-highlighting is done it does not make
sense to apply any highlighting while isearch is active.

Otherwise its almost impossible to see the matched area.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -66,6 +66,12 @@ _zsh_highlight()
   setopt localoptions noksharrays
   local REPLY # don't leak $REPLY into global scope
 
+  # Do not highlight in isearch widgets.
+  if [[ $WIDGET == zle-isearch-update ]]; then
+    region_highlight=()
+    return $ret
+  fi
+
   # Do not highlight if there are more than 300 chars in the buffer. It's most
   # likely a pasted command or a huge list of files in that case..
   [[ -n ${ZSH_HIGHLIGHT_MAXLENGTH:-} ]] && [[ $#BUFFER -gt $ZSH_HIGHLIGHT_MAXLENGTH ]] && return $ret


danielshahaf:
Can you describe how to reproduce the problem, starting from `zsh -f`?

Does this depend on some of your other PRs?  I guess it might depend on the second part of #255.

m0vie:
One example is this:
````
zsh -f
source zsh-syntax-highlighting.zsh
echo foo
print bar
vim<^R>
```
then type `echo`.

This results in 3 characters in the beginning behing highlighted in green, because `vim` was 3 characters long and the highlighting was not cleared. When nothing is typed before hitting ^R there is no (visible) issue.

danielshahaf:
Thanks, I can reproduce this.

As in https://github.com/zsh-users/zsh-syntax-highlighting/pull/259#issuecomment-168474898: if it's possible to write a test case for this it'd be great, but I realize that might be non-trivial.

> As long as zle_highlight for the isearch region can't be applied properly _after_ zsh-syntax-highlighting is done it does not make sense to apply any highlighting while isearch is active.

For future reference, the way to do this would be to mimic [the handling of zle_highlight[region] and zle_highlight[paste] in the driver](https://github.com/zsh-users/zsh-syntax-highlighting/blob/9b64ad750f356f828c3cfa96bc42489b2f081d8a/zsh-syntax-highlighting.zsh#L115-L142).

I think you can use `[[ $KEYMAP == 'isearch' ]]` in place of $YANK_ACTIVE and $REGION_ACTIVE.  I'm not sure off the top of my head how to get the start and end of the underlined segment.

… and I guess you might want to set zle_highlight+=(isearch:standout) if it's to be applied _on top of_ normal syntax highlighting.

m0vie:
> I think you can use [[ $KEYMAP == 'isearch' ]] in place of $YANK_ACTIVE and $REGION_ACTIVE. I'm not sure off the top of my head how to get the start and end of the underlined segment.

I looked into the zsh source, but sadly the beginning and end of the underlined segment are not exposed in the environment at all.

m0vie:
As in #259, this patch is also useless if `zle-isearch-update` is not bound before loading z-s-h.

In a clean `zsh -f`environment z-s-h does not get called at all when entering isearch and while typing in there. So this patch is the wrong approach.

danielshahaf:
> I looked into the zsh source, but sadly the beginning and end of the underlined segment are not exposed in the environment at all.

Expose them to widgets, then?  YANK_ACTIVE was only added in 5.1.1, specifically because z-sy-h needed it...

danielshahaf:
> As in #259, this patch is also useless if `zle-isearch-update` is not bound before loading z-s-h.
> 
> In a clean `zsh -f`environment z-s-h does not get called at all when entering isearch and while typing in there. So this patch is the wrong approach.

Is there some other thing we can do?  For example, could we do `if [[ $KEYMAP == isearch ]] ; then /* return without highlighting anything */; fi`?  (As it's better to highlight nothing than to highlight wrongly)

danielshahaf:
I can no longer reproduce this.  Could you please provide reproduction instructions again?  The instructions [above][1] don't reproduce the problem for me with current HEAD, even if I define a `zle-isearch-update` widget beforehand.

In the reproduction recipe, what widget does `^R` invoke?  The default keymap of `zsh -f` depends on $EDITOR, so it's not clear whether the keypress would be interpreted by the `emacs` or `viins` keymap.  (I tried both but neither reproduced the problem.)

[1]: https://github.com/zsh-users/zsh-syntax-highlighting/pull/257#issuecomment-168394665

m0vie:
I can still reproduce the issue using the same instructions.
`^R` invokes `history-incremental-search-backward`.

By the way: The fact that `zle-isearch-update`is defined does not *cause* the issue. Instead the fact that this "fix" works, relies on its definition.

danielshahaf:
So can I; I missed that I had to type `echo` after <kbd>C-r</kbd>.  Sorry for the noise.  I'll take the liberty of adding `bindkey -e` to the instructions.

danielshahaf:
Status:

1. Workaround: define zle-isearch-update before loading z-sy-h.
2. zsh-5.3 should invoke redrawhook (#245) from the isearch logic (see https://github.com/zsh-users/zsh-syntax-highlighting/pull/259#issuecomment-196044315), which should fix this issue, _if_ we will be able to tell when isearch mode is/isn't active.

m0vie:
About the workaround for zsh <5.3: If zle-isearch-update is not bound when z-sy-h is loaded, we could simply bind it to `_zsh_syntax_highlighting`.

danielshahaf:
For zsh that includes zsh-users/zsh@cbc44bd64a39649b7c73123c6eb85559d8d26f6c (workers/38145, due to be released in zsh-5.3), this was fixed by #283 / 79e4d3d1240599ab00a5e9aae54dfbd5c18fb270.

A different fix is needed for older zsh.

PR #258, opened 2015-12-31
highlight all available zsh glob pattern (not only * and ?)

m0vie:
Expansions are never analyzed for glob patterns to work around problems where complex constructs would wrongly be highlighted as glob patterns.


danielshahaf:
First of all, this changeset fixes a bug whereby `!*` is highlighted as [globbing] rather than as [history-expansion].

Could you please write tests?  The harness supports "expected failure" tests (printed in yellow/orange in the `make test` output) so you can also have tests for syntaxes you don't yet handle correctly.  (And you can write tests for the behaviour you're adding and we can merge them before we actually merge the implementation of that behaviour.)

I'll have to read the new function more carefully.

m0vie:
Updated the commit.

- Moved glob check before redirect to be able to highlight patterns like `<1-5>`.
- Fixed a typo for that check: `<[0*9]*-[0-9]*>` -> `<[0-9]*-[0-9]*>`
- Added tests for all covered cases
- Added two Xfail-tests for parameter expansions

m0vie:
Rebased onto latest master for the changes to the tests.

nicoulaj:

PR #261, opened 2015-12-31
use zle-line-pre-redraw hook if available

### LINKED ISSUE #245: redrawhook issues

m0vie:
The widget binding logic is kept, so that old versions of zsh still work as before.

_zsh_highlight_bind_widgets now accepts parameters, to specify which widgets to bind.

This is needed because with the redraw-hook we are not notifed for zle-isearch-* anymore, but we still need to react on those. Maybe this is an upstream issue?

TODO: The version numer in is-at-least actually needs to be 5.2.1, once officially released.

This fixes #245.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -262,14 +262,22 @@ _zsh_highlight_bind_widgets()
   local -U widgets_to_bind
   widgets_to_bind=(${${(k)widgets}:#(.*|run-help|which-command|beep|set-local-history|yank)})
 
-  # Always wrap special zle-line-finish widget. This is needed to decide if the
-  # current line ends and special highlighting logic needs to be applied.
-  # E.g. remove cursor imprint, don't highlight partial paths, ...
-  widgets_to_bind+=(zle-line-finish)
+  # Use zle-line-pre-redraw if available. Otherwise try binding widgets as fallback.
+  autoload -Uz is-at-least
+  if is-at-least 5.2; then
+    widgets_to_bind=(zle-line-pre-redraw)
+  else
+    widgets_to_bind=(${${(f)"$(builtin zle -la)"}:#(.*|orig-*|run-help|which-command|beep|set-local-history|yank)})
 
-  # Always wrap special zle-isearch-update widget to be notified of updates in isearch.
-  # This is needed because we need to disable highlighting in that case.
-  widgets_to_bind+=(zle-isearch-update)
+    # Always wrap special zle-line-finish widget. This is needed to decide if the
+    # current line ends and special highlighting logic needs to be applied.
+    # E.g. remove cursor imprint, don't highlight partial paths, ...
+    widgets_to_bind+=(zle-line-finish)
+
+    # Always wrap special zle-isearch-update widget to be notified of updates in isearch.
+    # This is needed because we need to disable highlighting in that case.
+    widgets_to_bind+=(zle-isearch-update)
+  fi
 
   local cur_widget
   for cur_widget in $widgets_to_bind; do


danielshahaf:
I have my own version of that at https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:mikachu-redrawhook/v2

Differences from your version:

- No special-casing of zle-isearch-update zle-isearch-exit
- Different ways of detecting support for the new hook (solve the is-at-least issue; btw the next release will be [called](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=37445) 5.3)
- Different error handling

danielshahaf:
> This is needed because with the redraw-hook we are not notifed for zle-isearch-* anymore, but we still need to react on those. Maybe this is an upstream issue?

Possibly.  A zle-isearch-update widget that modifies BUFFER doesn't trigger zle-line-pre-redraw.  Raise this subissue on -workers@?

danielshahaf:
@m0vie Ping, would you raise the `zle-line-pre-redraw` and isearch issue on -workers, please?  I think we should have that discussion before that hook is released by zsh.

m0vie:
I talked to Mikael on IRC and the he said the code was moved to the zle input loop for performance reasons to avoid multiple triggers. Hoewever the input loop is not triggered while minibuffers are active.

I raised the issue on -workers@ again.

danielshahaf:
> I raised the issue on -workers@ again.

At: http://www.zsh.org/mla/workers/2016/threads.html#00092 (workers/37557)

danielshahaf:
Once this is done, we can stop requiring that z-sy-h be the last thing sourced in zshrc.

danielshahaf:
... but we should document instead that z-sy-h registers a `zle-line-pre-redraw` hook, so if the user has such a hook (directly or via another plugin), he should arrange for his hook to call z-sy-h's hook.  (Upstream discussion: [workers/37639](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=37639)).

danielshahaf:
> Possibly. A zle-isearch-update widget that modifies BUFFER doesn't trigger zle-line-pre-redraw. Raise this subissue on -workers@?

This should have been fixed in zsh master, see https://github.com/zsh-users/zsh-syntax-highlighting/pull/259#issuecomment-196044315

m0vie:
Updated the pull request.
In case `zle-line-pre-redraw` is already bound, it is wrapped. In this case z-sy-h still needs to be loaded last, so I left the docs unchanged.

@danielshahaf I'm still using `is-at-least` for now. I looked at https://github.com/zsh-users/zsh-syntax-highlighting/commit/0cd9c4cbddf5cc18876cfaf1f60822a88a522571 and I don't understand how `.match-bracket` is related to this. For a current git version of zsh `zle -la .match-bracket` returns 1 for me.

danielshahaf:
> In case `zle-line-pre-redraw` is already bound, it is wrapped. In this case z-sy-h still needs to be loaded last, so I left the docs unchanged.

Okay.  I mentioned this in [workers/37639](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=37639).

About the use of `${(q)}` (https://github.com/zsh-users/zsh-syntax-highlighting/pull/261/files#diff-4ed73021a0fa2ff4e78a9909f4fc2f9bR292):

- Is zsh-syntax-highlighting.zsh idempotent?  Seems to me the code will happily do `f() { f; _zsh_highlight }` on the second `source zsh-syntax-highlighting.zsh` call, which will probably run into a recursion limit (or into the stack size limit, whichever comes first).

- Should there be validation that `[[ $currentbinding == user:* ]]` there?  Just belt and suspenders.

> @danielshahaf I'm still using `is-at-least` for now. I looked at https://github.com/zsh-users/zsh-syntax-highlighting/commit/0cd9c4cbddf5cc18876cfaf1f60822a88a522571 and I don't understand how `.match-bracket` is related to this. For a current git version of zsh `zle -la .match-bracket` returns 1 for me.

`match-bracket` is added by zsh-users/zsh@72f7f904e8734c2ee73b77e35e60ecb86148bc60 on the `mikachu/redrawhook` branch in zsh's repository, which the log message of 0cd9c4cbddf5cc18876cfaf1f60822a88a522571 references.  I was using the presence of that widget as a test for whether the redraw hook is supported by the zsh we run under.  (Should've added a comment to that effect.)

danielshahaf:
Typo: _zsh_highligh_pre_redraw_wrapper → _zsh_highlight_pre_redraw_wrapper

m0vie:
You're right, this does not work as intended. I refactored all my branches and I think I now have a solution that fixes all the recent issues we were having:

But I'm starting to get really confused with all those open pull requests and issues that depend on each other so I didn't submit yet another, yet.

I put everything here in a branch:
https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...m0vie:p_allzlefixes

Maybe we can talk on IRC in the next few days and to go through it so we can close some of those issues?

danielshahaf:
Status update: we spoke on IRC and @m0vie will submit the `p_allzlefixes` branch as a PR.  That PR would (IIUC) not depend on any other outstanding PR, so will be able to be reviewed/merged first.

danielshahaf:
@phy1729 points out that `is-at-least 5.2` should become `is-at-least 5.3`.

I'd like to merge one of the two versions of pre-redraw usage sooner: either just the first and fourth patches out of this PR, some variant of my version which I linked earlier.

m0vie:
Updated the PR. With the new `widgets_to_bind` logic, this is now much easier.

However the `is-at-least`condition needs to be changed to `5.3` we need to find a better way to decide if the feature is available.

m0vie:
Closing, since #356 is a better solution.

PR #266, opened 2016-01-15
region: Highlight the region even when it's inactive, e.g., after <set-mark-command>

danielshahaf:
Sometimes I type `echo hello <set-mark-command>world<quote-region>` (the default bindings are &lt;C-Space&gt; and &lt;M-"&gt; respectively).  This patch will cause the word `world` to be highlighted whilst its being typed.

Should an inactive region be highlighted differently from an active one?  Some widgets behave differently on active and inactive regions, e.g., pasting will delete the region only the region is active.  There are more differences, particularly in vi mode (grep Src/Zle/ for if/switch statements on the C variable `region_active`).

/cc @m0vie - this is in the same general area as #256 and #257

### DIFF
--- a/highlighters/main/test-data/mark.zsh
+++ b/highlighters/main/test-data/mark.zsh
@@ -1,0 +1,37 @@
+# -------------------------------------------------------------------------------------------------
+# Copyright (c) 2016 zsh-syntax-highlighting contributors
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are permitted
+# provided that the following conditions are met:
+#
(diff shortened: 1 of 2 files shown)


danielshahaf:
Conflicts due to 4849ef3307f37155b7d02294647d2ae011363392 in master.

danielshahaf:
The `MARK > 0` test is probably wrong: MARK and CURSOR are zero-based, if the mark is set before the first character of the buffer then `MARK` would legitimately be zero.

nicoulaj:

PR #273, opened 2016-02-17
use https instead of git for url

hackerkid:


danielshahaf:
What's the motivation for the change? To protect users from MITM attacks?

hackerkid:
Apart from that using git instead of https for git clone creates [problem](http://stackoverflow.com/questions/5860888/git-through-proxy) in systems which are using a proxy server for accessing the Internet. This can be a bit confusing. 

Other fix is to force git to use https instead of git

git config --global url.https://github.com/.insteadOf git://github.com/

danielshahaf:
Thanks for the additional information.  Merged.

Issue #224, opened 2015-10-29
Highlighting not applied during isearch


danielshahaf:
When recalling history with `history-incremental-search-backward` (`^R` in `bindkey -e` mode) — while the `isearch` keymap is active — the `$zle_highlight[isearch]` highlighting is applied, but no other highlighting is.

Normal syntax highlighting should take place during isearch.

Possibly relevant: [the `zle-isearch-update` widget](http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Special-Widgets); compare [Bart's email](www.zsh.org/cgi-bin/mla/redirect?USERNUMBER=18571) previously linked from #137.

danielshahaf:
Fixed in #283, requires zsh 5.3 (not yet released) for `$ISEARCH_ACTIVE`.

PR #283, opened 2016-03-28
apply suffix and isearch zle_highlights on top

m0vie:
Now that
`38145: ZLE parameters for isearch and completion suffixes`
got merged upstream, we can make use of these.


danielshahaf:
So, this PR fixes #224 and moots #257, is that correct?  How is this PR related to #259, I assume they are independent (both of them needed)?

danielshahaf:
In second thought, I think we'll still need to fix #257 separately older zsh.

m0vie:
Yes, this takes care of #224 and #257 for new versions of zsh. We can't do much for older versions, except disable highlighting in isearch altogether. #259 is completely unrelated. I was gonna close #259 because that does not have the effect I wanted it to have.

PR #285, opened 2016-04-01
tests: Fix grep syntax

phy1729:
In basic regular expressions | is not a special character that spearates
branches, so use an extended regular expression.

### DIFF
--- a/tests/test-highlighting.zsh
+++ b/tests/test-highlighting.zsh
@@ -148,7 +148,7 @@ fi
 # Process each test data file in test data directory.
 integer something_failed=0
 for data_file in ${0:h:h}/highlighters/$1/test-data/*.zsh; do
-  run_test "$data_file" | tee >($results_filter | ${0:A:h}/tap-colorizer.zsh) | grep -v '^not ok.*# TODO' | grep -q '^not ok\|^ok.*# TODO' && (( something_failed=1 ))
+  run_test "$data_file" | tee >($results_filter | ${0:A:h}/tap-colorizer.zsh) | grep -v '^not ok.*# TODO' | grep -Eq '^not ok|^ok.*# TODO' && (( something_failed=1 ))
   (( $pipestatus[1] )) && exit 2
 done
 


danielshahaf:
Since the code before the patch is equivalent to the code after the
patch, I take it the reason for the patch is not to fix a bug but to
make the code more readable.

+1 to commit.

phy1729:
Looking at what I think is the [relevant bit of POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03_02)

> The interpretation of an ordinary character preceded by a <backslash> ( '\\' ) is undefined, except for:
> - The characters ')', '(', '{', and '}'
> - The digits 1 to 9 inclusive (see BREs Matching Multiple Characters)
> - A character inside a bracket expression

OpenBSD's [re_format(7)](http://man.openbsd.org/re_format.7) states
> An ERE is one** or more non-empty** branches, separated by ‘|’.
> [...]
> Basic regular expressions differ in several respects:
> -  ‘|’, ‘+’, and ‘?’ are ordinary characters and there is no equivalent for their functionality.

Accordingly `grep '\|'` matches a literal pipe character. Now Ubuntu's [re_format(7)](http://manpages.ubuntu.com/manpages/trusty/en/man7/re_format.7.html) makes nearly the same statement
>  Obsolete ("basic") regular  expressions  differ  in  several  respects. '|',  '+',  and  '?' are ordinary characters and there is no equivalent for their functionality.

Yet, `grep '\|'` seems to match anything as if the pipe is separating empty branches.

This is all a long way to say the patch fixes the exit code of `make test` when there is a failing test when running on OpenBSD (and likely the other BSDs).

danielshahaf:
> Yet, `grep '\|'` seems to match anything as if the pipe is separating empty branches.

I see no contradiction here.  According to your references, POSIX states the meaning of `\|` is [undefined](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap01.html#tag_01_05_07); both re_format(7) manpages state the same; so treating `\|` as an alternation operator, or having the pattern `\|` match everything, are both compliant behaviours for an implementation.

> This is all a long way to say the patch fixes the exit code of `make test` when there is a failing test when running on OpenBSD (and likely the other BSDs).

So, the existing code relies on undefined behaviour, _and_ is broken, and you have a patch that fixes both issues?  Wonderful!  Please commit.

(I didn't know BRE's didn't support alternation... I hope there is some envvar I can set to make `\|` in a BRE become a fatal error, for development purposes)

PR #288, opened 2016-04-02
fixes for zle handling

m0vie:
New pull request replacing #257 and providing a proper solution for what #259 tried to do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,15 @@ custom widgets have been created (i.e., after all `zle -N` calls and after
 running `compinit`).  Widgets created later will work, but will not update the
 syntax highlighting.
 
+### Why does syntax highlighting not work while searching history?
+
+In `zsh` versions before 5.3 is not possible for `zsh-syntax-highlighting.zsh`
+to know if an incremental search is currently active and that matched parts of the
+buffer should be underlined (or otherwise highlighted). Therefore, it is not possible
+for `zsh-syntax-highlighting.zsh` to apply syntax highlighting and to underline the
+matched part of the search. While searching the history, the latter is more important,
+so syntax highlighting is disabled in this case.
+
 ### How are new releases announced?
 
 There is currently no "push" announcements channel.  However, the following
--- a/highlighters/cursor/cursor-highlighter.zsh
+++ b/highlighters/cursor/cursor-highlighter.zsh
@@ -34,15 +34,14 @@
 # Whether the cursor highlighter should be called or not.
 _zsh_highlight_cursor_highlighter_predicate()
 {
-  # accept-* may trigger removal of cursor highlighting
-  [[ $WIDGET == accept-* ]] ||
-    _zsh_highlight_cursor_moved
+  # remove cursor highlighting when the line is finished
+  [[ $WIDGET == zle-line-finish ]] || _zsh_highlight_cursor_moved
 }
 
 # Cursor highlighting function.
 _zsh_highlight_cursor_highlighter()
 {
-  [[ $WIDGET == accept-* ]] && return
+  [[ $WIDGET == zle-line-finish ]] && return
   
   _zsh_highlight_add_highlight $CURSOR $(( $CURSOR + 1 )) cursor
 }
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -62,9 +62,8 @@
 # Whether the highlighter should be called or not.
 _zsh_highlight_main_highlighter_predicate()
 {
-  # accept-* may trigger removal of path_prefix highlighting
-  [[ $WIDGET == accept-* ]] ||
-    _zsh_highlight_buffer_modified
+  # may need to remove path_prefix highlighting when the line ends
+  [[ $WIDGET == zle-line-finish ]] || _zsh_highlight_buffer_modified
 }
 
 # Helper to deal with tokens crossing line boundaries.
@@ -595,7 +594,7 @@ _zsh_highlight_main_highlighter_check_path()
 
   # If this word ends the buffer, check if it's the prefix of a valid path.
   if [[ ${BUFFER[1]} != "-" && ${#BUFFER} == $end_pos ]] &&
-     [[ $WIDGET != accept-* ]]; then
+     [[ $WIDGET != zle-line-finish ]]; then
     local -a tmp
     tmp=( ${expanded_path}*(N) )
     (( $#tmp > 0 )) && REPLY=path_prefix && return 0
--- a/highlighters/main/test-data/path_prefix2.zsh
+++ b/highlighters/main/test-data/path_prefix2.zsh
@@ -31,7 +31,7 @@
 # Related to path_prefix.zsh
 
 BUFFER='ls /bin/s'
-WIDGET=accept-line
+WIDGET=zle-line-finish
 
 expected_region_highlight=(
   "4 9 default"    # /bin/s
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -58,6 +58,14 @@ _zsh_highlight()
   # Store the previous command return code to restore it whatever happens.
   local ret=$?
 
+  # Remove all highlighting in isearch, so that only the underlining done by zsh itself remains.
+  # For details see FAQ entry 'Why does syntax highlighting not work while searching history?'
+  # and http://www.zsh.org/mla/users/2016/msg00415.html.
+  if [[ $WIDGET == zle-isearch-update ]]; then
+    region_highlight=()
+    return $ret
+  fi
+
   setopt localoptions warncreateglobal
   setopt localoptions noksharrays
   local REPLY # don't leak $REPLY into global scope
@@ -242,6 +250,10 @@ _zsh_highlight_bind_widgets()
   # current line ends and special highlighting logic needs to be applied.
   # E.g. remove cursor imprint, don't highlight partial paths, ...
   widgets_to_bind+=(zle-line-finish)
+
+  # Always wrap special zle-isearch-update widget to be notified of updates in isearch.
+  # This is needed because we need to disable highlighting in that case.
+  widgets_to_bind+=(zle-isearch-update)
 
   local cur_widget
   for cur_widget in $widgets_to_bind; do
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -237,6 +237,12 @@ _zsh_highlight_bind_widgets()
   # Override ZLE widgets to make them invoke _zsh_highlight.
   local -U widgets_to_bind
   widgets_to_bind=(${${(k)widgets}:#(.*|orig-*|run-help|which-command|beep|set-local-history|yank)})
+
+  # Always wrap special zle-line-finish widget. This is needed to decide if the
+  # current line ends and special highlighting logic needs to be applied.
+  # E.g. remove cursor imprint, don't highlight partial paths, ...
+  widgets_to_bind+=(zle-line-finish)
+
   local cur_widget
   for cur_widget in $widgets_to_bind; do
     case $widgets[$cur_widget] in
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -235,8 +235,10 @@ _zsh_highlight_bind_widgets()
   }
 
   # Override ZLE widgets to make them invoke _zsh_highlight.
+  local -U widgets_to_bind
+  widgets_to_bind=(${${(k)widgets}:#(.*|orig-*|run-help|which-command|beep|set-local-history|yank)})
   local cur_widget
-  for cur_widget in ${${(k)widgets}:#(.*|orig-*|run-help|which-command|beep|set-local-history|yank)}; do
+  for cur_widget in $widgets_to_bind; do
     case $widgets[$cur_widget] in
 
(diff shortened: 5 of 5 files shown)


danielshahaf:
Should `_zsh_highlight_set_or_wrap_special_zle_widget` do argument validation?  For example, check `[[ -n $1 ]]`, or possibly even `[[ $1 == zle-* ]]`?

Stylistically, I prefer `(( ${+widgets[$hook]} ))` to `[[ -z … ]]`.

troubleshooting.md: please link to [the relevant section of zshzle(1)](http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Special-Widgets).  Since `zle-*` aren't often added, I would tend to enumerate the current set of such widgets (`zle-isearch-exit` `zle-isearch-update` `zle-line-init` `zle-line-finish` `zle-history-line-set` `zle-keymap-select`), for greppability.

docs:

- troubleshooting.md: Please strike "This leads to some things that have to be considered and some issues": that sentence is not informative.

- The `zle-*` entry is unclear.  (It is clear with the context of our IRC conversation in mind; however, readers won't have that context.)  I suggest the following to clarify it: First, state the behaviour (widgets named `zle-*` do not trigger highlighting, except zle-line-finish and zle-isearch-update); then, give the rationale for it; and finally, address the edge case consequence (user-defined widgets matching that naming pattern won't be wrapped).

- isearch entry: The title is inaccurate: the "minibuffer" is the line below the prompt — the one that execute-named-cmd uses — and it never gets highlighted.  The issue is with highlighting the command shown in lieu of $BUFFER whilst an isearch is active.

- isearch entry: The body of the entry doesn't describe the failure mode of https://github.com/zsh-users/zsh-syntax-highlighting/pull/257#issuecomment-168394665.

- isearch entry: "Therefore" should be followed by a comma.

- isearch entry: s/it is not possible to apply highlighting by z-sy-h/it is not possible for z-sy-h to highlight/

- log message typo: "Old version" → "Old versions"

I haven't tested the fifth commit yet.

m0vie:
> Should `_zsh_highlight_set_or_wrap_special_zle_widget` do argument validation? For example, check `[[ -n $1 ]]`, or possibly even `[[ $1 == zle-* ]]`?

I don't think that this is necessary. The function is only used internally, and even with a check for `zle-*`, invalid input is still possible.

> Stylistically, I prefer `(( ${+widgets[$hook]} ))` to `[[ -z … ]]`.

I do, too. But most of `z-sy-h` uses `[[ -n` and `[[ -z` so I stuck with that.

> `zsh-*` exclusion

I dropped the whole commit excluding `zle-*` widgets by default. After thinking about it more, there is no reason to:
- If the user has bound some of those hooks already, it is likely that he does something to `BUFFER` in there. We should respect that and simply wrap the widget just as we did before.
- I initially added the exclusion to avoid binding a widget twice. But that can't happen due to the `_zsh_highlight_widget_*` check, so there is no need for special handling.

So I also rolled back the `troubleshooting.md` extraction.

> isearch entry: The title is inaccurate: the "minibuffer" is the line below the prompt — the one that execute-named-cmd uses — and it never gets highlighted. The issue is with highlighting the command shown in lieu of $BUFFER whilst an isearch is active.
> isearch entry: "Therefore" should be followed by a comma.
> isearch entry: s/it is not possible to apply highlighting by z-sy-h/it is not possible for z-sy-h to highlight/

I improved the wording of that paragraph (which can now be found in the FAQ in `README.md`).

> log message typo: "Old version" → "Old versions"

Fixed.

> isearch entry: The body of the entry doesn't describe the failure mode of #257 (comment).

The failure from that comment does not apply anymore now that we always bind `zle-isearch-update`. It can't happen anymore that highlighting from a previous line gets "stuck". The only remaining issue is that in old versions of `zsh` we can't do both -- apply syntax highlighting *and* underlining.

m0vie:
Rebased.

m0vie:
Changed the last commit about isearch. Removed the zsh-5.3-specific fallback. We can track that separately.

danielshahaf:
Review of current branch (6fbeac34d7808d9b3e8414540cecc65d88c0e34f):

1. Log messages: Please capitalize the first letter of the first line.

2. > I don't think that this is necessary. The function is only used internally, and even with a check for `zle-*`, invalid input is still possible.

    "Being used internally" is a weak argument: bugs can be internal too.  However, I think we do need to check this, since otherwise passing no argument will result in the first `if` branch being entered, and `zle -N` would be invoked with invalid arguments.

    I'm voting for a `${+widgets[$1]}` check.

3. > I dropped the whole commit excluding zle-* widgets by default. After thinking about it more, there is no reason to:

    I agree that `_zsh_highlight_set_or_wrap_special_zle_widget` needn't restrict itself to `zle-*` widgets only.  However, in that case, please update its docstring comment.

    Could you please push the old branch (the one I reviewed in my previous comment) somewhere?  I don't have it any more locally, and I'd like to interdiff the previous / current branches and to make sense of my previous review comments.

4. First commit: What if `[[ $currentbinding == completion:* ]]`?  Don't you need to do what `_zsh_highlight_bind_widgets` does?  In fact, shouldn't `+_zsh_highlight_set_or_wrap_special_zle_widget` and the body of the for loop in `_zsh_highlight_bind_widgets` share code?

    I think we also need a catchall `else` branch if the widget exists and the prefix of `$widgets[$1]` isn't recognised, but sharing code with `_zsh_highlight_bind_widgets` will fix this.

5. Second commit: I'm convinced that it's correct and would be happy to merge it as soon as the issues in its parent are addressed.

6. Third (tip) commit: Don't say "current `zsh` versions"; name specific lower or upper bound version numbers.  (The word "current" will bitrot.)  Also, mention $ISEARCHMATCH_ACTIVE.

6. Third (tip) commit: I'm having trouble reconciling the two comments, the one that says "We don't highlighting in isearch" and the one that says "We hook to isearch to receive updates".  The first comment should point to the FAQ entry and the FAQ entry to a zsh-workers@ thread about the zsh API lacuna in question.  The second comment should state _why_ we hook isearch (not only that we do so; that is evident from the code), since apparently the purpose isn't for highlighting to run.

Sorry for the long delay — it is entirely my fault.  I look forward to merging the first two commits as soon as they're ready (even if the third commit might then need another iteration).

danielshahaf:
Checklist (for me to review pre-merge):

- [ ] Review points in my last comment
- [ ] Check https://github.com/zsh-users/zsh-syntax-highlighting/pull/257#issuecomment-168394665 works
- [ ] Check `: /usr/loca<accept-line>` removes highlighting
- [ ] Test everything with a `$ISEARCHMATCH_ACTIVE`-less zsh
- [ ] Test everything with a `$ISEARCHMATCH_ACTIVE`-ful zsh

m0vie:
> Could you please push the old branch (the one I reviewed in my previous comment) somewhere? I don't have it any more locally, and I'd like to interdiff the previous / current branches and to make sense of my previous review comments.

Old version can be found here: https://github.com/m0vie/zsh-syntax-highlighting/tree/p_allzlefixes_old



> 1) Log messages: Please capitalize the first letter of the first line.

Done.

> 6) Third (tip) commit: Don't say "current zsh versions"; name specific lower or upper bound version numbers. (The word "current" will bitrot.) Also, mention $ISEARCHMATCH_ACTIVE.
> 
> 7) Third (tip) commit: I'm having trouble reconciling the two comments, the one that says "We don't highlighting in isearch" and the one that says "We hook to isearch to receive updates". The first comment should point to the FAQ entry and the FAQ entry to a zsh-workers@ thread about the zsh API lacuna in question. The second comment should state why we hook isearch (not only that we do so; that is evident from the code), since apparently the purpose isn't for highlighting to run.

Done

> 2) 3) 4)

I agree that the code should be unified. I changed it in the latest version. Now, in `_zsh_highlight_bind_widgets` it is possible to specify additional hooks that should be bound, even if they are currently unbound. This way, all the code is re-used and the whole thing is much simpler now.

danielshahaf:
Sorry for the very long delay ☹

Review (of 56b81a9b66524ce7e99adbd8105e6b5bd2af6a1f):

1. Log messages: `s/'driver':/driver:/`

2. First commit: the `''` case does not imply that the widget is a `zle-*` widget:

        $ zsh -f
        % bindkey x foo
        % [[ $widgets[foo] == '' ]] && echo empty
        empty

    So, the comment doesn't match the code.  One or the other should be fixed.

    Also, the `eval` there is not needed.  It's needed in the other branches for the closure effect (hardcoding the current value of the calling scope's `$cur_widget` in the body of the defined function), but the function _body_ in this case uses no parameters of the calling scope.

3. Second commit (8dfe7b091015b468933a298e0781e7478aac7392) is the same as 2ba33bb1b034f4433b7237fb888da0c20141cba9 (parent of 6fbeac34d7808d9b3e8414540cecc65d88c0e34f), which I deemed ready to merge in the previous review iteration.

4. Third commit: please mention the X-Seq of the message you link to.  (Or alternatively use a http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=12345 URL.)

5. Third commit: aren't there some conditions under which syntax highlighting _does_ work during an incremental search?  If so, I think the FAQ entry should state them up front (first sentence), to avoid creating a bad impression for FAQ readers who haven't yet installed z-sy-h.

    That first sentence could be something like this:

    > While searching history with the `history-incremental-search-backward` widget (bound by default to <kbd>Ctrl+R</kbd> in zsh's `bindkey -e` keymap), syntax highlighting only works if […].

6. The FAQ entry says "5.3 is", however, when I just merged this branch and then #261 onto a temporary head (https://github.com/zsh-users/zsh-syntax-highlighting/issues/245#issuecomment-234505258, second bullet, last paragraph) I got a merge conflict because the other head had "v5.3 it is".  Leaving the `v` bikeshed aside, the `it` is correct, so it seems something got lost in one of the rebases somewhere.  Anyway, please re-add that `it`.

danielshahaf:
> Second commit (8dfe7b0) is the same as 2ba33bb (parent of 6fbeac3), which I deemed ready to merge in the previous review iteration.

Thinking ahead, that commit adds `zle-line-finish` wrapping.  And `zle-line-finish` is one of the widgets supported by `add-zle-hook-widget`.  So I think the #245 solutions (https://github.com/zsh-users/zsh-syntax-highlighting/issues/245#issuecomment-234505258) will want to treat `zle-line-finish` and `zle-line-pre-redraw` the same way; namely: use `add-zle-hook-widget` if available and the current custom approach otherwise.

I don't think this consideration affects this branch, though; it's just something the other branches will need to do after 8dfe7b0 is merged.

danielshahaf:
I've implemented the suggestions from my penultimate comment on a branch: https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:m0vie-i288-v2?expand=1

I dropped the link to the "API for zsh5.3 concerning ISEARCHMATCH_ACTIVE" thread since it didn't seem relevant to the comment next to it (the thread is about 5.3 behaviour but the code is about 5.2 support); did you mean to link to a different thread?

I dropped the explanation from the first log message since upstream now documents that (commit users/21752: zsh-users/zsh@18d676ff199d7ddc8c9ff3e18a4764ca1a4b25a4).

m0vie:
Thank you for taking care of it. I'm very busy lately and didn't have time to look at it.

I agree with all the changes you made.

Just two small typos left:

1)
"Under those version of zsh it is not possible for..."
-> "versions"

2)
"Bind to z-sy-h driectly."
-> "directly"

Should I update this PR again, or are you going to merge your changes?

danielshahaf:
Fixed both and force-pushed my branch.  Thanks for the review.

My plan is first of all to get `add-zle-hook-widget` sorted upstream ([workers/38933](http://www.zsh.org/mla/workers/2016/msg01469.html)), then this issue, then the redrawhook issues.  I posted my review checklist for this issue hereinabove.

Re updating the PR, feel free: `git push m0vie +danielsh/m0vie-i288-v2:p_allzlefixes` It'd be a good thing — special cases are the root of all evil.

Quick question, what do you think of limiting the new codepath https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:m0vie-i288-v2#diff-4ed73021a0fa2ff4e78a9909f4fc2f9bR299 to `zle-*` widgets?  That is, of making that codepath behave as it does now for `[[ $cur_widget == zle-* ]]` and as the `*` codepath otherwise.

danielshahaf:
@phy1729 If you have a moment I'd love a third pair of eyes over the upcoming FAQ entry: https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:m0vie-i288-v2#diff-04c6e90faac2675aa89e2176d2eec7d8R35  Thanks!

m0vie:
> Quick question, what do you think of limiting the new codepath master...danielshahaf:m0vie-i288-v2#diff-4ed73021a0fa2ff4e78a9909f4fc2f9bR299 to zle-* widgets? That is, of making that codepath behave as it does now for [[ $cur_widget == zle-* ]] and as the * codepath otherwise.

Sounds like a good idea to be on the safe side!

> Re updating the PR, feel free:

OK, I'll update it with your changes later.

danielshahaf:
> Sounds like a good idea to be on the safe side!

I wrote a quick draft of this and appended it to the branch.

> OK, I'll update it with your changes later.

Okay.  I've stopped making changes to the branch here; the mutex is yours.

m0vie:
Merged in 171a4eec2d47a9c69ca26832c26e5278c3454442

danielshahaf:
And after the merge 4ad311e was committed, q.v..

PR #292, opened 2016-04-16
Adjust plugin.zsh file to run on zsh 5.1 in mSYS2.

adamkruszewski:
A sole filename caused "command not found" errors while loading with zplug or oh-no-zsh on zsh 5.1.1 inside mSYS2 (Windows).
Sourcing the main file seems to work fine.

### DIFF
--- a/zsh-syntax-highlighting.plugin.zsh
+++ b/zsh-syntax-highlighting.plugin.zsh
@@ -1,1 +1,1 @@
-source $(dirname $0)/zsh-syntax-highlighting.zsh
+source $0:A:h/zsh-syntax-highlighting.zsh
--- a/zsh-syntax-highlighting.plugin.zsh
+++ b/zsh-syntax-highlighting.plugin.zsh
@@ -1,1 +1,1 @@
-zsh-syntax-highlighting.zsh
+source $(dirname $0)/zsh-syntax-highlighting.zsh


phy1729:
Does

    source $0:A:h/zsh-syntax-highlighting.zsh

also work?

adamkruszewski:
Yes.

danielshahaf:
That needs braces (`${0:A:h}`), see 8f19af6b319db6130f376f3f2a27b5b74acd0562.

danielshahaf:
The underlying problem is probably that your system renders symlinks as plain files whose contents are the target of the symlink.  Converting the symlink to a plain file would sidestep the problem but not fix the root cause.

I don't mind applying your patch (or rather, https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:i292-v1, which also converts the *.plugin.zsh file to from a symlink to a plain file), but I would advise you to look into enabling symlinks in your platform, if possible...

adamkruszewski:
That would explain the problem. mSYS2 gives a linux like env on Windows, but lacks symlinks in any real meaning of the word :-/

luca020400:
zsh-syntax-highlighting/zsh-syntax-highlighting.zsh:1: job table full or recursion limit exceeded
Reverting this commit fixes it

danielshahaf:
`source zsh-syntax-highlighting.plugin.zsh` works for me in `zsh -f`.

`source zsh-syntax-highlighting.zsh` also works for me in `zsh -f`.

@luca020400 Can you show `ls -l` of these two files?  You could conceivably get the error if z-sy-h.zsh were a symlink to z-sy-h.plugin.zsh on your system....

luca020400:
Hmm for some reasons source zsh-syntax-highlighting.plugin.zsh overwrote source zsh-syntax-highlighting.zsh

PR #293, opened 2016-04-22
remove hook on sourcing again

seletskiy:


danielshahaf:
The hook function is currently idempotent, so I assume the purpose of this change is to avoid polluting $preexec_functions, and forward compatibility if the hook is made non-idempotent in the future, right?  Or does having the hook twice in $preexec_functions have any other undesirable effect?

seletskiy:
@danielshahaf: I often re-source single session, so I want to evade possible slow down after couple of `source`'s.

danielshahaf:
Do you actually get duplicates in $preexec_functions without this patch?  add-zsh-hook has been idempotent since zsh-4.3.10: https://github.com/zsh-users/zsh/commit/5e7e797ca90cd1c1b7ab54d39118f0b7469f348f#diff-de493ab4d85538b175a018ea4fb7d7d5

```zsh
% add-zsh-hook precmd _precmd_vcs_info
% add-zsh-hook precmd _precmd_vcs_info
% add-zsh-hook precmd _precmd_vcs_info
% typeset -p precmd_functions
typeset -a precmd_functions=( _precmd_vcs_info _precmd_read_comp )
%
```

seletskiy:
Hmm, looks like I was wrong on that one, sorry. That PR can be closed.

danielshahaf:
No problem.

PR #294, opened 2016-04-22
play well with zsh-autosuggest

seletskiy:


danielshahaf:
What does this fix?  What is the failure mode of the incumbent code?  I'm trying to understand what the problematic interaction is and whether it should be fixed in z-sy-h, in zsh-autosuggestions, or possibly in zsh itself.

I'm not convinced this patch is correct: what should we do in the case where z-sy-h is sourced, wraps some builtin widget, then the user wraps that widget and calls z-sy-h again?  E.g., `source zsh-syntax-highlighting.zsh; f() { zle .$WIDGET -w -- "$@" }; zle -N self-insert f; source zsh-syntax-highlighting.zsh`?  Should the second z-sy-h invocation wrap the (now user-defined) `self-insert` widget?

/cc @phy1729 @m0vie WDYT?

seletskiy:
@danielshahaf: I use both zsh-sy-h and `zsh-sug`. `zsh-sy-h` will wrap widgets, and then `zsh-sug` will wrap wrapped widgets. It's OK.

Then, when I re-source `~/.zshrc`, `zsh-sy-h` will tries to wrap wrapped by `zsh-sug` widgets (that are already wrapped by `zsh-sy-h`!). In that case, my zsh version is just dumps the core.

Regarding mentioned issue, I guess it's better to document, that user can define own widgets only before inclusion of `zsh-sy-h`, and not after.

danielshahaf:
> @danielshahaf: I use both `zsh-sy-h` and `zsh-sug`. `zsh-sy-h` will wrap widgets, and then `zsh-sug` will wrap wrapped widgets. It's OK.
> 
> Then, when I re-source ~/.zshrc, `zsh-sy-h` will tries to wrap wrapped by `zsh-sug` widgets (that are already wrapped by `zsh-sy-h`!). In that case, my zsh version is just dumps the core.
> 

Ah!  So it's a mutual recursion issue.  Thanks for getting to the bottom of it!

I wonder why zsh's safety check doesn't trigger:

```zsh
% f() { g }; g() { f }; f 
f:1: maximum nested function level reached
% zle -N f; bindkey "^T" f
% <press Ctrl+t>
f:2: maximum nested function level reached
%
```

Could you try these commands and report whether they give the error message I get or dump core?  I'm running zsh git HEAD-ish (5.3-to-be) on Linux, but I don't recall any fixes in the area since zsh-5.2 (which was reported to produce coredumps in this case).

> Regarding mentioned issue, I guess it's better to document, that user can define own widgets only before inclusion of `zsh-sy-h`, and not after.

It's already documented:
https://github.com/zsh-users/zsh-syntax-highlighting/#why-must-zsh-syntax-highlightingzsh-be-sourced-at-the-end-of-the-zshrc-file

However, documentation aside, we should better handle the use-case `repeat 2 { source zsh-autosuggestions; source z-sy-h }`.  At least, we shouldn't crash; ideally, highlighting would remain enabled, too.

My first priority is to add some hotfix to stop the crashes (even if that means reduced functionality for the rest of the affected session).  Perhaps this?

```diff
diff --git a/zsh-syntax-highlighting.zsh b/zsh-syntax-highlighting.zsh
index 9ff3a7f..462d9f2 100644
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -27,6 +27,7 @@
 # vim: ft=zsh sw=2 ts=2 et
 # -------------------------------------------------------------------------------------------------
 
+if ! (( ${+functions[_zsh_highlight]} )); then
 
 if [[ -o function_argzero ]]; then
   # $0 is reliable
@@ -312,3 +321,5 @@ add-zsh-hook preexec _zsh_highlight_preexec_hook 2>/dev/null || {
 
 # Initialize the array of active highlighters if needed.
 [[ $#ZSH_HIGHLIGHT_HIGHLIGHTERS -eq 0 ]] && ZSH_HIGHLIGHT_HIGHLIGHTERS=(main) || true
+
+fi
```

Once a hotfix has been applied, we can figure out how to DTRT in the "repeated `source`" use-case, too.

Makes sense?

danielshahaf:
Is the problem symmetric?  That is, do both `repeat 2 { source zsh-autosuggestions; source z-sy-h }` and `repeat 2 { source z-sy-h; source zsh-autosuggestions }` reproduce the error?

For future reference, zsh-autosuggestions' wrapping code seems to be: https://github.com/zsh-users/zsh-autosuggestions/blob/87facd9b85630f288433aa0a20a963cffc612ee5/zsh-autosuggestions.zsh#L116-L156

danielshahaf:
Hmm.  [Case (D) in zsh-autosuggestions#126 (Comment)](https://github.com/zsh-users/zsh-autosuggestions/issues/126#issuecomment-188460186) suggests it's _not_ a mutual recursion issue...

seletskiy:
Yep. This PR solves that problem.

On Fri, Apr 22, 2016 at 5:38 PM, Daniel Shahaf <notifications@github.com>
wrote:

> Hmm. Case (D) in zsh-autosuggestions#126 (Comment)
> <https://github.com/zsh-users/zsh-autosuggestions/issues/126#issuecomment-188460186>
> suggests it's *not* a mutual recursion issue...
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/zsh-users/zsh-syntax-highlighting/pull/294#issuecomment-213389275>
>



-- 
Stanislav Seletskiy

danielshahaf:
> I wonder why zsh's safety check doesn't trigger:
> 
> ```zsh
> % f() { g }; g() { f }; f 
> f:1: maximum nested function level reached
> % zle -N f; bindkey "^T" f
> % <press Ctrl+t>
> f:2: maximum nested function level reached
> %
> ```

I saw this warning in a `--enable-zsh-debug` build of latest zsh master, while an identical release build segfaulted.

The maximum depth is controlled by the `--enable-max-function-depth` configure switch.

seletskiy:
@danielshahaf: So, what about my PR? It effectively solves the issue.

danielshahaf:
@seletskiy As I said in my first reply, I think this PR breaks valid use-cases.  I think the right fix to the issue is https://github.com/zsh-users/zsh-autosuggestions/issues/126#issuecomment-217041261, and no change on z-sy-h's side will be required.

seletskiy:
@danielshahaf: It will not break valid use cases if user will specify plugins in the right order.

seletskiy:
@danielshahaf: And in your opinion, crash dump is better, than breaking some theoretical use cases, that should be used by someone right now to be broken?

danielshahaf:
> @danielshahaf: And in your opinion, crash dump is better, then breaking some theoretical use cases, that should be used by someone right now to be broken?

My first priority is to stop the segfault from occurring.  At this point I'm not sure a change to z-sy-h would be needed at all (Eric Freese wouldn't have seen my proposed z-asug patch yet).  Even if a change to z-sy-h would be needed, it would likely take the form not of this PR but of some other fix, since your PR introduces technical debt (as explained above).

seletskiy:
@danielshahaf: I'm not arguing, that you should merge *my* pull-request, I'm all for any better solution.

BUT. You seems to think, that issue should be solved in 3rd-party plugin (no offence intented). Don't you think, that sourcing of zsh-s-h should be idempotent? If I will have another plugin, that wraps some widgets, it will lead to crash dump again, and the only way to fix it is to go to the that new plugin source code and fix it.

So, in my opinion, zsh-s-h should be fixed as well as zsh-au.

danielshahaf:
> Don't you think, that sourcing of zsh-s-h should be idempotent?

Consider this case:

```zsh
% source zsh-syntax-highlighting.zsh 
% speaking-self-insert() { zle -M "You typed ${(qq)KEYS}"; zle .$WIDGET -- "$@" } 
% zle -N self-insert speaking-self-insert
% source zsh-syntax-highlighting.zsh
% source zsh-syntax-highlighting.zsh
```

Currently, the z-sy-h on line 4 happily wraps the new widget defined on line 3.  The invocations on lines 1 and 4 do different things, _because the shell state has changed in the meantime_.  On the other hand, the invocation on line 5 is a no-op.

What should the `self-insert` widget be bound to after line 4?  Should it be bound to `.self-insert` (the builtin meaning), or to `speaking-self-insert`, or to the z-sy-h wrapper of one of those two?

> If I will have another plugin, that wraps some widgets, it will lead to crash dump again, and the only way to fix it is to go to the that new plugin source code and fix it

You seem to be arguing that all plugins in the world would have to change to accomodate z-sy-h.  That is not the case.

In fact, the opposite is true: the widget binding strategy used by z-sy-h generalizes arbitrarily.  You could install any number of plugins that wrap widgets in the manner z-sy-h does and it would work.  (z-sy-h uses the strategy described at https://github.com/zsh-users/zsh-autosuggestions/issues/126#issuecomment-217121315.)

I think the question is simply whether z-sy-h should implement loop prevention, right?

danielshahaf:
> In fact, the opposite is true: the widget binding strategy used by z-sy-h generalizes arbitrarily.  You could install any number of plugins that wrap widgets in the manner z-sy-h does and it would work.  (z-sy-h uses the strategy described at https://github.com/zsh-users/zsh-autosuggestions/issues/126#issuecomment-217121315.)

When I wrote this I was thinking of the common use-case, of each plugin being sourced exactly once.  Supporting something like `plugins=( foo bar foo baz foo qux )` would require a trivial change (have `foo` use unique widget names every time it's sourced).

danielshahaf:
Closing, since all the remaining work is tracked by other issues (#245 here and https://github.com/zsh-users/zsh-autosuggestions/issues/126#issuecomment-217041261; the former requires zsh 5.3 but the latter works with any zsh).

PR #301, opened 2016-04-24
Use ZSH_HIGHLIGHT_STYLES keys

### LINKED ISSUE #300: simplify code: work with the keys of ZSH_HIGHLIGHT_STYLES until assignment
All the code currently works like this:
````
style=$ZSH_HIGHLIGHT_STYLES[path]
...
zle_highlight+=("start end $style")
````

The following would be simpler, more readable, would carry more semantics (for styles with identical value), and could potentially simplify testing a lot:

````
style=path
...
zle_highlight+=("start end $ZSH_HIGHLIGHT_STYLES[$style]")
````

phy1729:
Closes #300. This changes the interface for the patten highlighter.

### DIFF
--- a/highlighters/brackets/brackets-highlighter.zsh
+++ b/highlighters/brackets/brackets-highlighter.zsh
@@ -83,10 +83,10 @@ _zsh_highlight_brackets_highlighter()
       local bracket_color_size=${#ZSH_HIGHLIGHT_STYLES[(I)bracket-level-*]}
       local bracket_color_level=bracket-level-$(( (levelpos[$pos] - 1) % bracket_color_size + 1 ))
       local style=$bracket_color_level
-      region_highlight+=("$pos $((pos + 1)) $ZSH_HIGHLIGHT_STYLES[$style]")
+      _zsh_highlight_add_highlight $pos $((pos + 1)) $style
     else
       local style=bracket-error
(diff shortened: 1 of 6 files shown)


danielshahaf:
Thank you for doing the legwork!

Looks good.  Two minor comments:

- About the backwards-incompatible change to the pattern highlighter: could we make the transition smoother?  For example, perhaps the pattern highlighter could do:

    ```diff
    -    _zsh_highlight_add_highlight $((mbegin[2] - 1)) $mend[2] $ZSH_HIGHLIGHT_PATTERNS[$pat]
    +    if (( $+ZSH_HIGHLIGHT_STYLES[ZSH_HIGHLIGHT_PATTERNS[$pat]] )); then
    +      _zsh_highlight_add_highlight $((mbegin[2] - 1)) $mend[2] $ZSH_HIGHLIGHT_PATTERNS[$pat]
    +    else
    +      # remove $pat from $ZSH_HIGHLIGHT_PATTERNS and warn to stderr
    +    fi
    ```

    And we definitely need a warning at the top of changelog (or in NEWS or release notes or equivalent) about this change.

- The second commit will cause a textual (non-semantic) conflict with @m0vie's #298, in the `case` branches.

Modulo these two issues, I think this is ready to merge :-)

Issue #306, opened 2016-05-05
Document conventions for third-party highlighter authors


danielshahaf:
As the subject says.  Including:

- [x] Style names should be prefixed by the highlighter name (e.g., `foo` highlighter uses `foo-*` keys)

ascii-soup:
Happy to pick this up, will try and do this weekend.

PR #309, opened 2016-05-09
docs: Added style key prefix convention

ascii-soup:
Part of #306, just added the prefix/namespace to the docs.


danielshahaf:
Looks good.  The git commit doesn't include your email address in the author/committer metadata; would you like to fix this before merging?

Will merge this as soon as you reply.

PR #314, opened 2016-05-12
main-highlighter can be run as command

psprint:
I've added 4 blocks of code that make main-highlighter runnable as a command. This sounds bad, but I think it isn't. It's your call :)

### DIFF
--- a/highlighters/main/parse.zsh
+++ b/highlighters/main/parse.zsh
@@ -1,0 +1,16 @@
+#!/bin/zsh
+
+source "../../zsh-syntax-highlighting.zsh"
+
+if [ -n "$1" ]; then
+    # Load from given file
+    PREBUFFER=""
+    BUFFER="$(<$1)"
+
+    _zsh_highlight_main_highlighter
+
+    # This output can be diffed to detect changes in operation 
+    print -rl -- "${region_highlight[@]}"
+else
+    echo "Usage: ./parse.zsh {file}"
+fi


danielshahaf:
Thanks, but I think it'd be cleaner to do it the other way around: to keep `main-highlighter.zsh` as a library file and have a new file do `source main-highlighter.zsh` and then call `_zsh_highlight_main_highlighter`?

`print -rl $foo` is an antipattern; there should be a `--` guard there: `print -rl -- $foo`.

psprint:
Updated. Much better architecture arose. No changes to existing code is needed, the additional file `parse.zsh` covers all.

danielshahaf:
- We can't assume `/bin/zsh` exists; that's a linuxism, upstream's default is `/usr/local/bin/zsh`.  We could use `#!/usr/bin/env zsh` but then we can't also pass `-f` on the shebang line; or we could use `#!/usr/local/bin/zsh -f` but then it won't work on linux out of the box.

    Can we have our cake and it eat too?  That is, can we both use a `zsh` from PATH _and_ invoke it with `-f`?

- Line 3: Please don't assume cwd is highlighters/main/, i.e., use a `${0:A:h}/../../foo` style path.  (The driver needs to beware of nofunctionargzero because it's sourced, but I think that's not an issue for an executed script.)

- Line 5: It would be better to use `[ -f ]`, `[ -r ]`, or `BUFFER="$(<$1)" || { echo ...; exit 1 }`, in ascending order of preference.

- Should `make install` install parse.zsh?  I guess not?  In that case, it should probably live in some other directory (such as highlighters/../tools/).

danielshahaf:
Ping?

psprint:
* /bin/zsh is also on OS X, but in general you're right, although: this would be used only by developers so fixing shebang isn't really a problem, and you could have the -f there this way. Besides, developers would probably have multiple Zsh versions installed and they would tune shebang anyway

* How would the ${0:A:h} path look like?

* That's doable

* We could install parse.zsh to have a debugging tool equipped for users. Maybe sometimes it would serve as component of debugging instructions, that would be given e.g. in some Issue conversation

danielshahaf:
1. Yes, this is a dev-only tool, but that's not an excuse to leave the job half done.  How about setting `#!/bin/sh` and having `if [ -z "$ZSH_VERSION" ]; then exec /usr/bin/env zsh -f "$@" ; fi` as the first line?

2. `${0:A:h}/main-highlighter.zsh`.  (The braces are required in case ksharrays is set.)

3. *nod*

4. I think I wasn't clear: right now parse.zsh _does_ get installed by `make install`, but I think that was an unintended consequence.  Moving parse.zsh out of highlighters/main/ is one way to prevent it from being installed.

danielshahaf:
Continued as #314.

PR #315, opened 2016-05-12
Optimization that in my tests shows 2.7s -> 2.2s

psprint:
New approach to what #312 tried to do. Nicely catches how text pointers (start_pos, end_pos) are advanced and advances proc_buf by the same amount. Runs in 2.2s.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -199,6 +199,8 @@ _zsh_highlight_main_highlighter()
   #
   local this_word=':start:' next_word
   integer in_redirection
+  # Processing buffer
+  local proc_buf="$buf"
   for arg in ${interactive_comments-${(z)buf}} \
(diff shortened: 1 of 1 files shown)


danielshahaf:
LGTM.  Could you update the new block comment to use full sentences and punctuation (full stop at the end of a sentence, blank lines between paragraphs, etc)?

psprint:
Is it ok now? BTW, I tried moving proc_buf advancement to the end – where your patch located it, and test still ran in 2.2s. I'm sure the tests are correct, wonder how is it possible that your version produces the same output, but in different time.

danielshahaf:
Merged in 5a98cd3.

> BTW, I tried moving proc_buf advancement to the end – where your patch located it, and test still ran in 2.2s. I'm sure the tests are correct, wonder how is it possible that your version produces the same output, but in different time.

Let's have any investigation of this issue happen o na separate ticket; I'd like to close this one.

PR #317, opened 2016-05-12
'main': 2nd optimization that in my tests shows 2.2s -> 2.0s

psprint:
Switch from [,-1] to [,len] gives speed improvement.


danielshahaf:
Merged in ef4bfe5 with comment changes.

PR #320, opened 2016-05-13
'main': Implement simple command type cache

m0vie:
Improves performance a lot when editing complex command lines that would lead to many `type -w` calls, each performing a `PATH` search.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -89,6 +89,12 @@ _zsh_highlight_main_add_region_highlight() {
 #
 # The result will be stored in REPLY.
 _zsh_highlight_main__type() {
+  if (( $+_zsh_highlight_main__command_type_cache )); then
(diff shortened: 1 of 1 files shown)


danielshahaf:
What happens when add-zsh-hook is unavailable?  (Either because using a pre-4.3.4 zsh, or because autoload failed for some other reasons.)

Proceeding without the cache would be fine.  A comment to that effect beside the error message would be useful.  However, the global associative array would then never be cleared.

Since shells might be very long-running (think a `screen` session that runs for years), I think we do need to ensure we never let the global array grow without being cleared.  Perhaps by only defining the global variable if installing the hook succeeded, and then checking its existence (`${+foo}`) before assigning to it?

Otherwise I'd be happy to merge this as soon as I have sufficient time.

m0vie:
It would indeed be bad if the array never got cleared. And I agree, better to just disable the cache altogether in the unlikely case that add-zsh-hook is not available.

I updated the pull request with your suggestions.

danielshahaf:
Thanks!

PR #326, opened 2016-05-27
'main': Internal cleanup: drop $style_override.

danielshahaf:
@m0vie points out that there's a simpler alternative to using `$style_override`.


danielshahaf:
This is related to #260.

danielshahaf:
(Ignore that b92ced7 — its log message incorrectly referred to this issue instead of to issue 316.)

PR #328, opened 2016-06-02
'main': Fix highlighting of «\~», which isn't a filename expansion.

danielshahaf:
Add use of the `(e)` parameter flag, inspired by https://github.com/zsh-users/zsh/blob/770500d69db0cf9dcee6ed62455aa79a38ed182e/Completion/Unix/Type/_path_files#L257-L268.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -687,7 +687,11 @@ _zsh_highlight_main_highlighter_expand_path()
   # The $~1 syntax normally performs filename generation, but not when it's on the right-hand side of ${x:=y}.
   setopt localoptions nonomatch
   unset REPLY
-  : ${REPLY:=${(Q)${(e)~1}}}
+  if eval 'REPLY=${(e)~1}' 2>/dev/null; then
+    REPLY=${(Q)REPLY}
+  else
+    REPLY=$1
+  fi
 }
 
 # -------------------------------------------------------------------------------------------------
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -687,7 +687,7 @@ _zsh_highlight_main_highlighter_expand_path()
   # The $~1 syntax normally performs filename generation, but not when it's on the right-hand side of ${x:=y}.
   setopt localoptions nonomatch
   unset REPLY
-  : ${REPLY:=${(Q)~1}}
+  : ${REPLY:=${(Q)${(e)~1}}}
 }
 
 # -------------------------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


danielshahaf:
Summary of relevant zshall(1) information:

- Order of stages:
  - History Expansion;
  - Alias Expansion;
  - Process Substitution, Parameter Expansion, Command Substitution, Arithmetic Expansion, Brace Expansion;
  - Filename Expansion;
  - Filename Generation.

- Meaning of flags:
  - `(Q)`:      Remove one level of quotes from the resulting words.

  - `(e)`:      Perform parameter expansion, command substitution and arithmetic expansion on the result. Such expansions can be nested but too deep recursion may have unpredictable effects.
  - `~`:        Treat any characters resulting from parameter expansion as being eligible for filename expansion and filename generation. [However, filename generation is not performed on the right-hand side of ${x:=y}.]

In what order should these three be applied, then?  (There may be more than 3! options because of parameter expansion idiosyncrasies.) Is the permutation this PR uses correct?

danielshahaf:
Bug: typing in `${PWD}.exe` aborts after the `self-insert` of `{` with `_zsh_highlight_main_highlighter_expand_path:7: bad substitution`.  However, if `${PWD}.exe` is pasted in one word, it is highlighted correctly (more correctly than in master, in fact: see the just-added test).

danielshahaf:
The pertinent part seems to be this:

> `(e)`: Perform parameter expansion, **command substitution** and arithmetic expansion on the result. Such expansions can be nested but too deep recursion may have unpredictable effects.

This means `: $(echo /)` will now highlight the entire command substitution as a `[path]`... but to do so the highlighting code will *execute* that command substitution *every character typed*.

This idea needs further thought.

nicoulaj:

phy1729:
I don't think we should execute any part of the $BUFFER to decide highlighting (we have discussed this before, and I think we agree). Currently `ls \~` highlights the `\~` as path which I think is better than default. Perhaps cherry-pick path-dollared-word4.zsh and an amended path-tilde-home3.zsh and close?

danielshahaf:
> I don't think we should execute any part of the $BUFFER to decide highlighting (we have discussed this before, and I think we agree).

+1

> Could we just `${(Q)${~1}}`?

The incumbent code uses `${(Q)~1}`.  What's the difference between the two?

phy1729:
It forces quote removal to happen after filename expansion.
```
a=\\~
print -r ${(Q)~a} ${(Q)${~a}}
/home/phy1729 ~
```

danielshahaf:
Matthew Martin wrote on Tue, 12 Dec 2017 06:21 +0000:
> It forces quote removal to happen after filename expansion.
> ```
> a=\\~
> print -r ${(Q)~a} ${(Q)${~a}}
> /home/phy1729 ~
> ```

This seems correct, what are you waiting for? :)

phy1729:
A second set of eyes just to be sure. :)

danielshahaf:
Is a regression test needed? tests/generate.zsh

phy1729:
Rarely a bad idea to add a test, but I didn't want to steal your commit since you already have the test in your PR.

PR #334, opened 2016-06-04
Use sed to activate the plugin with Oh My ZSH in INSTALL.md

danieliancu:
Use sed to activate the plugin with Oh My ZSH in INSTALL.md.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -60,7 +60,7 @@ your `.zshrc`.
 
 2. Activate the plugin in `~/.zshrc`:
 
-        plugins=( [plugins...] zsh-syntax-highlighting)
+        sed -i "s/\(plugins=(.*\))/\1 zsh-syntax-highlighting)/" ~/.zshrc
 
 3. Source `~/.zshrc`  to take changes into account:
 


danielshahaf:
> `+        sed -i "s/\(plugins=(.*\))/\1 zsh-syntax-highlighting)/" ~/.zshrc`

This won't work, for several reasons:

1. The `-i` flag is not portable.  (On BSD it takes an argument, and on other systems it might not exist at all.)
2. The meaning of `\1` inside double quotes differs between bash, dash, and zsh, so the command given won't work if run from dash.  (Using single quotes instead of double quotes would address this.)
3. sedding an arbitrary source code robustly is _very_ hard.  For example, as written this will mangle assignments to variables named `earplugins` too.  It will also match inside string literals, fail to handle multiline array assignments, handle two assignments on a single line wrongly, and so on...

Frankly, I think the best solution here is for oh-my-zsh to provide an "install a plugin" command that a user can run from an interactive shell to register a plugin.  That hypothetical "oh-my-zsh-register-and-enable-plugin" command (named better) could then take care of updating ~/.zshrc.  (It could use a guard comment to delimit the file portion being autoedited to avoid the third problem.)

Thanks for the patch!

danieliancu:
@danielshahaf Thank you very much for the comment and this awesome package.

Will work on it.

danielshahaf:
You're welcome, but don't forget to thank @nicoulaj who founded the project :-)

danielshahaf:
Closing due to inactivity, feel free to reopen if you have an update.

PR #355, opened 2016-08-24
'main': Work around type -w bug in zsh

### LINKED ISSUE #354: Bug: broken completion
Hi there,

[This commit from the 28th of March](https://github.com/zsh-users/zsh-syntax-highlighting/commit/2f18ba64e397f11b04e685eeaee802b7285510cc) seems to introduce a problem with completion, at least on my system.

The problem is that if I am typing the name of a command in the current working directory, including a leading `./`, completion picks up names of commands in other folders in my `$PATH`.

For example, let's say I'm doing some Django development and I want to run `./manage.py`.  Then, if I type `./man` and hit tab, it completes to `./man ` (with a trailing space), presumably because it's finding `man` in my `$PATH`. Same for `ls`, `sudo`, `yes`, `wc`, etc.

My understanding/expectation is that given the leading `./`, it should *only* complete to something in the cwd, and ignore the rest of `$PATH`.

I've stripped my `~/.zshrc` back to this single line and I still see the problem:

```
source $HOME/prog/others/zsh-users/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
```

Sadly I don't have the zsh knowledge to attempt to fix this, so I'm raising it as an issue. As a temporary workaround, I've rolled my local copy of `zsh-syntax-highlighting` back to [the previous commit](https://github.com/zsh-users/zsh-syntax-highlighting/commit/3ce01076b521d9ab37dce90f2e73bdf1d44ec3d7), which seems to work fine.

Other info:

```
ld% echo $PATH    
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games

ld% zsh --version
zsh 5.0.2 (x86_64-pc-linux-gnu)

ld% lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.5 LTS
Release:	14.04
Codename:	trusty
```

m0vie:
Even if 'type -w' returns 'none', it still hashes it in $commands:
http://www.zsh.org/mla/workers/2016/msg01583.html

Before 12b879caf7a6f96921d1c03ab9b839a7f6c8eb1e, we ran 'type -w'
in a subshell, so this bug did not manifest. Now that we don't,
the $commands array is poisoned with invalid entries.

To prevent this, while still keeping the perfomance benefit of
avoiding the subshell, we make sure we remove invalid entries
from $commands again.

Fixes #354.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -101,7 +101,13 @@ _zsh_highlight_main__type() {
   fi
   unset REPLY
   if zmodload -e zsh/parameter; then
-    if (( $+aliases[(e)$1] )); then
+    if ! builtin type -w -- $1 >/dev/null 2>&1; then
+      REPLY=none
+      # work around zsh bug: even if type -w encounters an invalid command,
+      # it hashes it anyways. http://www.zsh.org/mla/workers/2016/msg01583.html
+      # we force an unhash here to keep the hash tables clean.
+      (( $+commands[(e)$1] )) && unhash $1
+    elif (( $+aliases[(e)$1] )); then
       REPLY=alias
     elif (( $+saliases[(e)${1##*.}] )); then
       REPLY='suffix alias'
@@ -113,8 +119,6 @@ _zsh_highlight_main__type() {
       REPLY=builtin
     elif (( $+commands[(e)$1] )); then
       REPLY=command
-    elif ! builtin type -w -- $1 >/dev/null 2>&1; then
-      REPLY=none
     fi
   fi
   if ! (( $+REPLY )); then


danielshahaf:
Alternative approach: avoid the bug, rather than clean up after it:

```diff
diff --git a/highlighters/main/main-highlighter.zsh b/highlighters/main/main-highlighter.zsh
index b120bf2..e1ebe6b 100644
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -81,6 +81,8 @@ _zsh_highlight_main_add_region_highlight() {
   _zsh_highlight_add_highlight $start $end "$@"
 }
 
+autoload -U is-at-least
+
 # Get the type of a command.
 #
 # Uses the zsh/parameter module if available to avoid forks, and a
@@ -113,7 +115,8 @@ _zsh_highlight_main__type() {
       REPLY=builtin
     elif (( $+commands[(e)$1] )); then
       REPLY=command
-    elif ! builtin type -w -- $1 >/dev/null 2>&1; then
+    elif { is-at-least 5.3 || [[ /$1/ != *(/./|/.//)* ]] } &&
+         ! builtin type -w -- $1 >/dev/null 2>&1; then
       REPLY=none
     fi
   fi
```

Considerations:

- performance hit on cygwin for `.` or `..` command words (due to falling through to the fork)
- situations where the cache poison is an absolute path
- situations where the cache poison was added by the PATH_DIRS option

TODO:
- Swap the disjuncts, the pattern match is faster (h/t @m0vie).  (Or: just do 'is-at-least' once at global scope, as though a C static)

danielshahaf:
Applied a slightly different patch in 51614ca.

PR #356, opened 2016-08-24
Use add-zle-hook-widget zle-line-pre-redraw

### LINKED ISSUE #245: redrawhook issues

danielshahaf:
Fixes #245, q.v..

Should be rebased/squashed before merging.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -43,7 +43,9 @@ if true; then
 fi
 
 integer zsh_highlight_use_redrawhook
-if zle -la .match-bracket; then
+if autoload -U +X -- add-zle-hook-widget 2>/dev/null;
+   [[ "${${(@f)"$(which -- add-zle-hook-widget)"}[2]}" != $'\t'$histchars[3]' undefined' ]];
+then
   (( zsh_highlight_use_redrawhook=1 ))
 fi
 
@@ -275,7 +277,7 @@ _zsh_highlight_bind_widgets()
 
 if (( $zsh_highlight_use_redrawhook )); then
   _zsh_highlight_bind_widgets(){}
-  zle -N zle-line-pre-redraw _zsh_highlight
+  add-zle-hook-widget zle-line-pre-redraw _zsh_highlight
 fi
 
 # Load highlighters from directory.
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -42,6 +42,25 @@ if true; then
   fi
 fi
 
+# This function takes a single argument F and returns True iff F is an autoload stub.
+_zsh_highlight__function_is_autoload_stub_p() {
+  if (( ${+functions} )); then
+    ## zsh/parameter is available
+    #(( ${+functions[$1]} )) &&
+    [[ "$functions[$1]" == *"builtin autoload -X" ]]
+  else
+    ## zsh/parameter isn't available
+    #[[ $(type -wa -- "$1") == *'function'* ]] &&
+    [[ "${${(@f)"$(which -- "$1")"}[2]}" == $'\t'$histchars[3]' undefined' ]]
+  fi
+  # Do nothing here: return the exit code of the if.
+}
+
+# Return True iff the argument denotes a function name.
+_zsh_highlight__is_function_p() {
+  (( ${+functions[$1]} )) || [[ $(type -wa -- "$1") == *'function'* ]]
+}
+
 # This function takes a single argument F and returns True iff F denotes the
 # name of a callable function.  A function is callable if it is fully defined
 # or if it is marked for autoloading and autoloading it at the first call to it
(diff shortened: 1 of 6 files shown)


zsh-syntax-highlighting.zsh
```diff
@@ -42,6 +42,13 @@ if true; then
   fi
 fi
 
+integer zsh_highlight_use_redrawhook
+if autoload -U +X -- add-zle-hook-widget 2>/dev/null;
+   [[ "${${(@f)"$(which -- add-zle-hook-widget)"}[2]}" != $'\t'$histchars[3]' undefined' ]];
+then
+  (( zsh_highlight_use_redrawhook=1 ))
+fi
+
```
m0vie: Do we need this section? Instead in ---- SETUP ---- we could simply do:

if "add-zle-hook-widget is availabe"; then
  "add-zle-hook-widget"
else
  "bind widgets"
fi

That would ready clearer, imho.

zsh-syntax-highlighting.zsh
```diff
@@ -268,6 +275,22 @@ _zsh_highlight_bind_widgets()
...
+    # Reset $WIDGET since the 'main' highlighter depends on it.
+    #
+    # A nested function is required to hide zle parameters; see 
+    # "User-defined widgets" in zshall.
+    () {
+      local -h +r WIDGET=zle-line-finish
+      _zsh_highlight "$@"
+    } "$@"
+  }
+  _zsh_highlight_bind_widgets(){}
```
m0vie: This line could be removed if the comment above was implemented.

danielshahaf:
Merging onto current master (51614ca2c994486dc0a41ec0737a97f0af16a505), the three tests trigger the following warnings:

- highlighters/brackets/test-data/cursor-matchingbracket.zsh
- highlighters/main/test-data/alias-assignment1.zsh
- highlighters/pattern/test-data/rm-rf.zsh

```
% make -s quiet-test ZSH=$prefix/zsh/bin/zsh
Running test brackets
add-zle-hook-widget:21: job table full or recursion limit exceeded
_zsh_highlight:39: command not found: _zsh_highlight_highlighter_brackets_predicate
Bail out! output on stderr
# cursor-matchingbracket
1..1
not ok 1 - [9,9] «)» - expected "cursor-matchingbracket", observed "".
Running test main
add-zle-hook-widget:21: job table full or recursion limit exceeded
_zsh_highlight:39: command not found: _zsh_highlight_highlighter_main_predicate
Bail out! output on stderr
# alias-assignment1
1..1
not ok 1 - [1,3] «x=y» - expected "unknown-token", observed "".
Running test pattern
add-zle-hook-widget:21: job table full or recursion limit exceeded
_zsh_highlight:39: command not found: _zsh_highlight_highlighter_pattern_predicate
Bail out! output on stderr
# rm-rf
1..1
not ok 1 - [1,8] «rm -rf /» - expected "fg=white,bold,bg=red", observed "".
Makefile:38: recipe for target 'test' failed
make[1]: *** [test] Error 2
Makefile:49: recipe for target 'quiet-test' failed
make: *** [quiet-test] Error 2
zsh: exit 2     make -s quiet-test ZSH=$prefix/zsh/bin/zsh
```

Adding `zmodload zsh/zle` to the top of tests/test-highlighting.zsh resolves this.  This appears to be an infinite loop in upstream's add-zle-hook-widget; will report upstream.

danielshahaf:
> This appears to be an infinite loop in upstream's add-zle-hook-widget; will report upstream.

[workers/39128](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=39128)

danielshahaf:
Upstream will fix this; however, a fix on this end is needed too:

```diff
diff --git tests/test-highlighting.zsh tests/test-highlighting.zsh
index 1d169d6..d164590 100755
--- tests/test-highlighting.zsh
+++ tests/test-highlighting.zsh
@@ -47,6 +47,9 @@
   exit 2
 }
 
+# Required for add-zle-hook-widget.
+zmodload zsh/zle
+
 # Load the main script.
 . ${0:h:h}/zsh-syntax-highlighting.zsh
 
diff --git zsh-syntax-highlighting.zsh zsh-syntax-highlighting.zsh
index 1c3e8e4..035f3ac 100644
--- zsh-syntax-highlighting.zsh
+++ zsh-syntax-highlighting.zsh
@@ -330,8 +330,8 @@ if (( $zsh_highlight_use_redrawhook )); then
     } "$@"
   }
   _zsh_highlight_bind_widgets(){}
-  add-zle-hook-widget zle-line-pre-redraw _zsh_highlight
-  add-zle-hook-widget zle-line-finish _zsh_highlight__zle-line-finish
+  add-zle-hook-widget zle-line-pre-redraw _zsh_highlight || return 42
+  add-zle-hook-widget zle-line-finish _zsh_highlight__zle-line-finish || return 42
 fi
 
 # Load highlighters from directory.
```

Will flesh this out and add it to the branch.

zsh-syntax-highlighting.zsh
danielshahaf: Thanks @m0vie, I made a similar change in 5591e2adb1f7.

danielshahaf:
(github doesn't point out 2cf6a48, which was a sync-merge from master.)

danielshahaf:
(I can't edit the log messages of earlier commits on the branch, so remarking here instead for posterity:)  The reference to `.match-bracket` in ef41c9232d62e1ae7cf2cadb9ce8a6372341dddf is a bug.  That test was meant to detect zsh≥5.3, but it was incorrect since that widget name never existed in zsh.

danielshahaf:
Merged and pushed to the `feature/redrawhook` branch.  Posted a [call for testing](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=39670) to -workers@.

danielshahaf:
(This issue should be reopened.)

danielshahaf:
Ready to merge, just changelog needs to be updated.

danielshahaf:
Also fixes #507.

phy1729:
https://travis-ci.org/zsh-users/zsh-syntax-highlighting/builds/441045121

Seems that 4.3.11 and 4.3.12 do not support passing arguments to an anonymous function. Perhaps we should drop the `"$@"`in `_zsh_highlight__zle-line-finish()` (since they aren't used anyway) to retain compatibility to 4.3.11.

danielshahaf:
> https://travis-ci.org/zsh-users/zsh-syntax-highlighting/builds/441045121
> 
> Seems that 4.3.11 and 4.3.12 do not support passing arguments to an 
> anonymous function. Perhaps we should drop the `"$@"`in 
> `_zsh_highlight__zle-line-finish()` (since they aren't used anyway) to 
> retain compatibility to 4.3.11.

"aren't used"?  Couldn't someone's zshrc be calling `zle zle-line-finish -- foo bar baz` from some custom widget or another?

Formally we require 4.3.17+ since before I had commit access, but I also remember a bug report from a 4.3.9 user.

The next version, 4.3.13, was released on 2011-11-30, i.e., seven years ago.

How about doing here what we do with `grep -q` elsewhere: use the new idiom, and revert to the old one if someone complains that their system doesn't support the new idiom?  Or we could just name the anonymous function.

phy1729:
Summary of IRC:
`:A` is in 4.3.10 (thanks @nicoulaj), so supporting 4.3.11+ shouldn't be an issue there. I'm a fan of keeping support for 4.3.11 until CentOS 6 is no longer supported (Nov 2020) (Debian wheezy is on 4.3.17 and seems to already be EOL).

Since `_zsh_highlight` does not use its arguments, it should be safe to drop `"$@"` from the anonymous function and the `_zsh_highlight` invocation.

PR #358, opened 2016-08-31
Declare global variables

penman:
This caused warnings with `setopt warn_create_global`


danielshahaf:
Ross Penman wrote on Wed, Aug 31, 2016 at 07:56:33 -0700:
> This caused warnings with `setopt warn_create_global`

Thanks, pushed with a log message tweak.

PR #371, opened 2016-09-25
main-highlighter can be run as command, with profiling equipped

psprint:
Following up to https://github.com/zsh-users/zsh-syntax-highlighting/pull/314 (I've once pruned Z-SY-H fork when cleaning up my repositories, so have to start a new PR).
1. Addressed this, #!/bin/sh is the shebang, zsh -f is being run
2. Using `$0` and also `%N`
3. Doing `[[ -r "$1" ]]`
4. I opt for shipping `mh-parse.zsh` as it can serve for possible problem-investigation instructions at user's machine.

### DIFF
--- a/highlighters/main/mh-parse.zsh
+++ b/highlighters/main/mh-parse.zsh
@@ -1,0 +1,62 @@
+#!/bin/sh
+
+#
+# This file runs main highlighter on a specified file
+# i.e. parses the file with the highlighter. Outputs
+# running time (stderr) and resulting region_highlight
+# (stdout).
+#
+
+# Strive for -f
+# .. occuring problems (no q flag), but at least try to quote $0 and $@
+[[ -z "$ZSH_VERSION" ]] && exec /usr/bin/env zsh -f -c "source \"$0\" \"$1\" \"$2\" \"$3\""
+
+ZERO="${(%):-%N}"
+ZERO_RESOLVED="${ZERO:A:h}"
+
+#
+# Load Z-SY-H
+#
+
+if [[ -e "${ZERO_RESOLVED}/main-highlighter.zsh" ]]; then
+    source "${ZERO_RESOLVED}/../../zsh-syntax-highlighting.zsh"
+elif [[ -e "${0:A:h}/main-highlighter.zsh" ]]; then
+    source "${0:A:h}/../../zsh-syntax-highlighting.zsh"
+elif [[ -e "../../main-highlighter.zsh" ]]; then
+    source "../../zsh-syntax-highlighting.zsh"
+else
+    print "Could not find zsh-syntax-highlighting.zsh, aborting"
+    exit 1
+fi
+
+#
+# Call _zsh_highlight_highlighter_main_paint
+#
+
+if [[ -r "$1" ]]; then
+    # Load from given file
+    PREBUFFER=""
+    BUFFER="$(<$1)"
+
+    typeset -F SECONDS
+    SECONDS=0
+
+    _zsh_highlight_highlighter_main_paint
+
+    print -u2 "Main highlighter's running time: $SECONDS"
+
+    # This output can be diffed to detect changes in operation
+    print -rl -- "${region_highlight[@]}"
+else
+    if [[ -z "$1" ]]; then
+        print -u2 "Usage: ./mh-parse.zsh {to-parse file}"
+        exit 1
+    else
+        print -u2 "Unreadable to-parse file \`$1', aborting"
+        exit 2
+    fi
+fi
+
+exit 0
+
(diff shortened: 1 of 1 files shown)


highlighters/main/mh-parse.zsh
```diff
@@ -0,0 +1,62 @@
...
+#
+# This file runs main highlighter on a specified file
+# i.e. parses the file with the highlighter. Outputs
+# running time (stderr) and resulting region_highlight
+# (stdout).
+#
+
+# Strive for -f
+# .. occuring problems (no q flag), but at least try to quote $0 and $@
+[[ -z "$ZSH_VERSION" ]] && exec /usr/bin/env zsh -f -c "source \"$0\" \"$1\" \"$2\" \"$3\""
```
danielshahaf: sh scripts can't use `[[`; change it to `[`.

danielshahaf: About quoting:   `/usr/bin/env zsh -fc 'source "$@"' - "$0" "$@"` should do the trick.

highlighters/main/mh-parse.zsh
```diff
@@ -0,0 +1,62 @@
...
+
+ZERO="${(%):-%N}"
+ZERO_RESOLVED="${ZERO:A:h}"
```
danielshahaf: You don't actually need $ZERO_RESOLVED; it'd be fine to replace all uses with $ZERO.

Yes, I know I said `${0:A:h}` originally.  I think I mislead you; sorry for that.

highlighters/main/mh-parse.zsh
```diff
@@ -0,0 +1,62 @@
...
+    # Load from given file
+    PREBUFFER=""
+    BUFFER="$(<$1)"
+
+    typeset -F SECONDS
+    SECONDS=0
+
+    _zsh_highlight_highlighter_main_paint
+
+    print -u2 "Main highlighter's running time: $SECONDS"
```
danielshahaf: I'd rather send this to stdout, since it's not an error.  Perhaps make it a `# comment` to ease parsing?

highlighters/main/mh-parse.zsh
```diff
@@ -0,0 +1,62 @@
...
+
+if [[ -r "$1" ]]; then
+    # Load from given file
+    PREBUFFER=""
+    BUFFER="$(<$1)"
+
+    typeset -F SECONDS
+    SECONDS=0
+
+    _zsh_highlight_highlighter_main_paint
```
danielshahaf: The code is in no way specific to the `main` highlighter; I think it would be better suited in tools/ and taking a highlighter name as an argument.  WDYT?

Like the existing test scripts, I think this script would benefit from the stub `_zsh_highlight_add_highlight` implementation used by various files in tests/.  (It'd cause the output to print style names rather than color settings derived from them.)

highlighters/main/mh-parse.zsh
```diff
@@ -0,0 +1,62 @@
...
+    print -u2 "Main highlighter's running time: $SECONDS"
+
+    # This output can be diffed to detect changes in operation
+    print -rl -- "${region_highlight[@]}"
+else
+    if [[ -z "$1" ]]; then
+        print -u2 "Usage: ./mh-parse.zsh {to-parse file}"
+        exit 1
+    else
+        print -u2 "Unreadable to-parse file \`$1', aborting"
```
danielshahaf: Two minor issues:

- Prefix $0 to the error message.
- Use ${(qq)1} to escape the filename.

highlighters/main/mh-parse.zsh
```diff
@@ -0,0 +1,62 @@
...
+
+    _zsh_highlight_highlighter_main_paint
+
+    print -u2 "Main highlighter's running time: $SECONDS"
+
+    # This output can be diffed to detect changes in operation
+    print -rl -- "${region_highlight[@]}"
+else
+    if [[ -z "$1" ]]; then
+        print -u2 "Usage: ./mh-parse.zsh {to-parse file}"
```
danielshahaf: I'd rather the argument read `{file to parse}`; I think that scans better in English.

highlighters/main/mh-parse.zsh
danielshahaf: Don't you need to `setopt interactivecomments` here, so that any comments in the file would be highlighted correctly?

highlighters/main/mh-parse.zsh
psprint: A quick reply: this allows to redirect region_highlight output to a file, and have running time on terminal

nicoulaj:

PR #372, opened 2016-09-25
'main' / *_check_path: Precompute $#BUFFER and use it for speed gain

psprint:
Running main highlighter on itself (both runs are on the optimized
version):

- clean (8 runs, last 3 noted):
1.7007670000
1.7330720000
1.7038810000

- optimized (8 runs, last 3 noted):
1.5007230000
1.5142960000
1.4973320000

Average difference: 0.208456

When parsing main-highlighter with itself, the *_check_path function is
called 426 times. Note that there are 686 region_highlight resulting
entries.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -212,6 +212,7 @@ _zsh_highlight_highlighter_main_paint()
   local -a options_to_set # used in callees
   local buf="$PREBUFFER$BUFFER"
   integer len="${#buf}"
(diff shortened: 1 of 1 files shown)


danielshahaf:
Thanks, merged.  Currently, the code computes `$#buf` and `$#PREBUFFER`, so if $PREBUFFER is large, it is walked twice.  I think it'd be cleaner to do

```zsh
integer pure_buf_len=$#PREBUFFER
integer len=$(( pure_buf_len + $#BUFFER ))
```

to avoid that.  When entering a multi-line command $PREBUFFER tends to be larger than $BUFFER; I think that's the scenario in which this optimisation is most likely to be efficacious.

PR #384, opened 2016-10-27
Add Fedora package

suraia:
zsh-syntax-highlighting is now available in Fedora.

For some reason, the package link does not work yet. The package is available already, though: https://bodhi.fedoraproject.org/updates/?packages=zsh-syntax-highlighting


danielshahaf:
Merged.  Thanks for packaging z-sy-h!

PR #385, opened 2016-11-02
'main': Fix bug: no start_pos=$end_pos in comment short path

psprint:
No start_pos advance in comment short path


danielshahaf:
Thanks for the patch @psprint.  I've added a regression test and pushed it.  For future reference, your PR was based on a somewhat old revision of master, although that didn't matter in this case.  Thanks again for the patch!

PR #395, opened 2016-12-01
driver: Don't undefine aliases that begin with a '+', to workaround an upstream bug

### LINKED ISSUE #392: Warning: (eval):1: bad option: -x
Recently, after update, I get this message:

```
(eval):1: bad option: -x
```

**Screenshot:** 

screen shot 2016-11-22 at 13 42 27

After researching a bit, I understood that this message is caused by zsh-syntax-highlighting plugin, so it's gone when I disable the plugin. I've even tried to fix the problem by myself, but I can't understand when exactly it appears.

danielshahaf:
Fixes #392.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -28,8 +28,12 @@
 # -------------------------------------------------------------------------------------------------
 
 # First of all, ensure predictable parsing.
-zsh_highlight__aliases=`builtin alias -L`
-builtin unalias -m '*'
+zsh_highlight__aliases=`builtin alias -Lm '[^+]*'`
+# In zsh <= 5.2, `alias -L` emits aliases that begin with a plus sign ('alias -- +foo=42')
+# them without a '--' guard, so they don't round trip.
+#
+# Hence, we exclude them from unaliasing:
+builtin unalias -m '[^+]*'
 
 # Set $0 to the expected value, regardless of functionargzero.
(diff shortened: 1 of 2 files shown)


danielshahaf:
This assumes aliases are emitted one per line (even if the alias or its expansion include newlines) and in sorted order.  That appears to be the case with 5.3, however, 5.0.7 does not escape newlines there.

zsh-syntax-highlighting.zsh
```diff
@@ -29,7 +29,19 @@
...
+# In zsh <= 5.2, `alias -L` emits aliases that begin with a plus sign ('alias -- +foo=42')
+# them without a '--' guard, so they don't round trip.
+#
+# Hence, we exclude them from unaliasing...
+builtin unalias -m '[^+]*'
+# ... and if any aliases remain --- they must be "+foo" aliases --- then we
+# arrange *not to restore* them, since that would result in "alias: bad option"
+# errors.
+if [[ -n $zsh_highlight__aliases ]]; then
+  zsh_highlight__aliases=`comm -23 =(<<<"$zsh_highlight__aliases") =(alias -L)`
```
phy1729: Could avoid shelling out to comm with
```
temp=${(f)"$(alias -L)"};
_zsh_highlight__aliases=${${(f)_zsh_highlight_aliases}:|temp};
```
or similar

danielshahaf:
Added regression test.

PR #399, opened 2016-12-15
Update zplug install instruction

ace03uec:
The latest version of zplug has deprecated the `nice` tag. The `defer` tag is the preferred method now.


danielshahaf:
Merged in 4f49c4a35f17 with a log message tweak (added "${area}:" prefix). Thanks!

PR #400, opened 2016-12-15
fix: duplicated slash

vintersnow:
It seems that the slash is duplicated at `"$highlighter_dir/${highlighter}-highlighter.zsh"`


danielshahaf:
> It seems that the slash is duplicated at `"$highlighter_dir/${highlighter}-highlighter.zsh"`

Does this have any ill effect beyond the slash being duplicated in the
error message the patch touches?

vintersnow:
I think it doesn't...

So, should I not change the error message?

danielshahaf:
Merged with a changed log message.  Thanks!

PR #403, opened 2016-12-22
Code and test changes for running with NO_UNSET, as per issue 398

paxe:
I tested this in all of 5.07, 5.2, and 5.3 -- test results look to be same as before, and quiet-test is silent. To get NO_UNSET I just added -u to the Makefile rather than alter test scripts for it.

I suspect this may have uncovered a related bug with 5.3 only; see the comments around line 313 in the zsh-syntax-highlighting.zsh file, and let me know if you figure out a better fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ test:
 	for test in highlighters/*; do \
 		if [ -d $$test/test-data ]; then \
 			echo "Running test $${test##*/}"; \
-			$(ZSH) -f tests/test-highlighting.zsh "$${test##*/}"; \
+			$(ZSH) -fu tests/test-highlighting.zsh "$${test##*/}"; \
 			: $$(( result |= $$? )); \
 		fi \
 	done; \
--- a/highlighters/brackets/brackets-highlighter.zsh
+++ b/highlighters/brackets/brackets-highlighter.zsh
@@ -40,7 +40,7 @@
 # Whether the brackets highlighter should be called or not.
 _zsh_highlight_highlighter_brackets_predicate()
 {
-  [[ $WIDGET == zle-line-finish ]] || _zsh_highlight_cursor_moved || _zsh_highlight_buffer_modified
+  [[ ${WIDGET-} == zle-line-finish ]] || _zsh_highlight_cursor_moved || _zsh_highlight_buffer_modified
 }
 
 # Brackets highlighting function.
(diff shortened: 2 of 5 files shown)


danielshahaf:
> I tested this in all of 5.07, 5.2, and 5.3 -- test results look to be
> same as before, and quiet-test is silent. To get NO_UNSET I just added
> -u to the Makefile rather than alter test scripts for it.

Thank you very much!

> I suspect this may have uncovered a related bug with 5.3 only; see the
> comments around line 313 in the zsh-syntax-highlighting.zsh file, and
> let me know if you figure out a better fix.

[For future reference, that's line 313 of b68c2701d016e9502904760fd7fd0657b25f2b08, which
is currently tip of the PR.)

I think the fix is two-fold:

1. Invoke add-zle-hook-widget inside an anonymous function that does
`setopt localoptions nounset`.  (That'll have to remain for
compatibility with zsh 5.3.1.)

2. Fix add-z-h-w upstream to work when nounset is set in the calling
scope.  We could start changing `${widgets[$hook]}` to
`${widgets[$hook]:-}` on that line 84, but it's probably easier to throw
an `emulate -L` at the top of the function.  If you work on this, you
should send the patches to zsh-workers@zsh.org.

> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/zsh-users/zsh-syntax-highlighting/pull/403

I'll review it when I can, however that won't be today, sorry.

I notice you've merged the redrawhook branch as part of your PR.  My
plan is to review your branch, and later decide how to merge things:
whether to merge redrawhook first and then this PR, or to merge this
PR and thereby close the redrawhook stuff as well.

(Probably the former, since the redrawhook work is enough of a spaghetti
as it is...)

Thanks again for the patchset!

Oh, and we're on #zsh-syntax-highlighting on freenode, if you like to say hello.

paxe:
I don't even recall whether it was intentional or accidental that I started from feature/redrawhook, as I did the initial fork a few months back. No matter, I can easily re-do for you elsewhere, there wasn't much to it... mostly testing to see where things broke.

nicoulaj:

phy1729:
Sorry I didn't notice this earlier. I've committed a few changes in #451 to address the same NO_UNSET issue which came up in #449. Hopefully this fixes you issues as well although I didn't change anything around add-zle-hook-widget. If you wouldn't mind testing current master and reporting back if you still have any issues, that would be appreciated. Thanks for the PR!

phy1729:
Closing since I think the changes mentioned above fix the issue.

Issue #412, opened 2017-01-25
zsh 5.4-to-be features in use


danielshahaf:
The following zsh 5.4-to-be behaviours are used in the code, and must not appear in a z-sy-h release until they have appeared in a zsh release:

- [x] ALIAS_FUNC_DEF option (9523d6d49cb3d4db5bd84c3cec6168a2057fe3ab)
- [x] #415 (which fixes #407) assumes [workers/40285](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=40285)
    - pushed in 462779629a0c
    - update references to "5.3.2" to "5.4" or "5.3.1.1"
    - if upstream reverts workers/40285, revert or adapt #415 fix.

danielshahaf:
As expected, neither of these features was changed upstream prior to upstream's tag.

PR #415, opened 2017-02-07
workaround for PAT_STATIC bug in zsh

### LINKED ISSUE #407: history-incremental-pattern-search-backward is broken after upgrading to zsh 5.3.1
I recently upgraded zsh from 5.2-2 -> 5.3.1-1, in Arch Linux. After upgrading, `history-incremental-pattern-search-backward` and `history-incremental-pattern-search-forward` no longer work. That is, they show the most recent match, then `failing bck-i-search: foo`, then one other match from eons ago, then it fails to find any more. Reverting to zsh 5.2-2 fixes this problem.

I use [zim](https://github.com/Eriner/zim). When I disable `zsh-syntax-highlighting`, pattern searching works again. I tried creating a new minimal zsh install, that did not source `zim` at all. As soon as I directly source `zsh-syntax-highlighting`, pattern searching breaks again. Unlike issue #387, I'm *not* using `zsh-autocorrect`.

docwhat:
ZSH versions less than 5.3.2 (or 5.4) have a bug that prevents
`history-incremental-pattern-search-backward` for working
correctly (the history stops searching after the first found item).

Closes #407

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,10 +46,13 @@ syntax highlighting.
 
 ### Does syntax highlighting work during incremental history search?
 
-Highlighting the command line during an incremental history search
-(with the `history-incremental-search-backward` widget, which is
-bound by default to <kbd>Ctrl+R</kbd> in zsh's emacs keymap) requires zsh 5.3
-or newer.
+Yes!
+
+If you're using `history-incremental-search-backward` (by default bound to <kbd>Ctrl+R</kbd>
+in zsh's emacs keymap) then it works with _zsh version 5.3 and newer_.
+
+If you're using `history-incremental-pattern-search-backward`, then syntax highlighting works
+in _zsh 5.3.2 and newer_ due to [a bug in zsh](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=40285).
 
 Under zsh 5.2 and older, the zsh-default [underlining][zshzle-Character-Highlighting]
 of the matched portion of the buffer remains available, but zsh-syntax-highlighting's
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -53,6 +53,15 @@ fi
 # Core highlighting update system
 # -------------------------------------------------------------------------------------------------
 
+# Use workaround for bug in ZSH?
+# zsh-users/zsh@48cadf4 http://www.zsh.org/mla/workers//2017/msg00034.html
+autoload -U is-at-least
+if is-at-least 5.3.2; then
+  zsh_highlight__pat_static_bug=false
+else
+  zsh_highlight__pat_static_bug=true
+fi
+
 # Array declaring active highlighters names.
 typeset -ga ZSH_HIGHLIGHT_HIGHLIGHTERS
 
@@ -67,7 +76,9 @@ _zsh_highlight()
 
   # Remove all highlighting in isearch, so that only the underlining done by zsh itself remains.
   # For details see FAQ entry 'Why does syntax highlighting not work while searching history?'.
-  if [[ $WIDGET == zle-isearch-update ]] && ! (( $+ISEARCHMATCH_ACTIVE )); then
+  # This disables highlighting during isearch (for reasons explained in README.md) unless zsh is new enough
+  # and doesn't have the 5.3.1 bug
+  if [[ $WIDGET == zle-isearch-update ]] && { $zsh_highlight__pat_static_bug || ! (( $+ISEARCHMATCH_ACTIVE )) }; then
     region_highlight=()
     return $ret
   fi
@@ -393,8 +404,6 @@ add-zsh-hook preexec _zsh_highlight_preexec_hook 2>/dev/null || {
 
 # Load zsh/parameter module if available
 zmodload zsh/parameter 2>/dev/null || true
-
-autoload -U is-at-least
 
 # Initialize the array of active highlighters if needed.
 [[ $#ZSH_HIGHLIGHT_HIGHLIGHTERS -eq 0 ]] && ZSH_HIGHLIGHT_HIGHLIGHTERS=(main)


m0vie:
Isn't `is-at-least` itself using patterns and therefore causing the exact same bug it tries to prevent?

We should probably determine and remember this beforehand when sourcing z-sy-h, also for performance reasons.

docwhat:
@m0vie as soon as I read your comment, I realized you had to be right. I double checked in my plain-vanilla "fake" config and my work-around didn't work.

I pushed up a revised version that stores the value and uses that.  It's also easier to read, which I like.

danielshahaf:
@m0vie Thanks for reviewing.

@docwhat Besides my other comment (just posted), I have only a few minor style tweaks.  Will go through them once the functionality is settled.

danielshahaf:
@docwhat Feel free to join us on #zsh-syntax-highlighting on freenode.

README.md
```diff
@@ -46,10 +46,13 @@ syntax highlighting.
...
-(with the `history-incremental-search-backward` widget, which is
-bound by default to <kbd>Ctrl+R</kbd> in zsh's emacs keymap) requires zsh 5.3
-or newer.
+Yes!
+
+If you're using `history-incremental-search-backward` (by default bound to <kbd>Ctrl_R</kbd>
+in zsh's emacs keymap) then it works with _zsh version 5.3 and newer_.
+
+If you're using `history-incremental-pattern-search-backward`, then syntax highlighting works
+in _zsh 5.3.2 and newer_ due to [a bug](http://www.zsh.org/mla/workers//2017/msg00034.html).
```
docwhat: Right now it describes what the code does...

* Will zsh ever release a 5.3.2?
* If so, would it just include security patches, or would it include this patch to?
* Should I just change the `is-at-least` to 5.4?

zsh-syntax-highlighting.zsh
```diff
@@ -53,6 +53,15 @@ fi
 # Core highlighting update system
 # -------------------------------------------------------------------------------------------------
 
+# Use workaround for bug in ZSH?
+# zsh-users/zsh@48cadf4 http://www.zsh.org/mla/workers//2017/msg00034.html
+autoload -U is-at-least
```
docwhat: done

zsh-syntax-highlighting.zsh
```diff
@@ -53,6 +53,15 @@ fi
...
+if is-at-least 5.3.2; then
+  zsh_highlight__pat_static_bug=f
+else
+  zsh_highlight__pat_static_bug=t
+fi
```
docwhat: done

zsh-syntax-highlighting.zsh
```diff
@@ -67,7 +76,7 @@ _zsh_highlight()
 
   # Remove all highlighting in isearch, so that only the underlining done by zsh itself remains.
   # For details see FAQ entry 'Why does syntax highlighting not work while searching history?'.
-  if [[ $WIDGET == zle-isearch-update ]] && ! (( $+ISEARCHMATCH_ACTIVE )); then
+  if [[ $WIDGET == zle-isearch-update ]] && ([[ $zsh_highlight__pat_static_bug == t ]] || ! (( $+ISEARCHMATCH_ACTIVE )) ); then
```
docwhat: done

README.md
```diff
@@ -46,10 +46,13 @@ syntax highlighting.
 
 ### Does syntax highlighting work during incremental history search?
 
-Highlighting the command line during an incremental history search
-(with the `history-incremental-search-backward` widget, which is
-bound by default to <kbd>Ctrl+R</kbd> in zsh's emacs keymap) requires zsh 5.3
-or newer.
+Yes!
+
+If you're using `history-incremental-search-backward` (by default bound to <kbd>Ctrl_R</kbd>
```
docwhat: typo on my part. :-/

docwhat:
Pulling this out for visibility...

@danielshahaf said:
> Could we reword this to reflect that 5.3.2 is not yet released (and will probably be called 5.4)? I'll update #412 to refer to this.

Right now it describes what the code does...

* Will zsh ever release a 5.3.2?
* If so, would it just include security patches, or would it include this patch to?
* Should I just change the `is-at-least` to 5.4?

README.md
```diff
@@ -46,10 +46,13 @@ syntax highlighting.
...
-(with the `history-incremental-search-backward` widget, which is
-bound by default to <kbd>Ctrl+R</kbd> in zsh's emacs keymap) requires zsh 5.3
-or newer.
+Yes!
+
+If you're using `history-incremental-search-backward` (by default bound to <kbd>Ctrl+R</kbd>
+in zsh's emacs keymap) then it works with _zsh version 5.3 and newer_.
+
+If you're using `history-incremental-pattern-search-backward`, then syntax highlighting works
+in _zsh 5.3.2 and newer_ due to [a bug in zsh](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=40285).
```
m0vie: `history-incremental-search-backward` or `history-incremental-pattern-search-backward` doesn't matter, we don't check for that. Its _5.3.2 and newer_ in all cases

zsh-syntax-highlighting.zsh
```diff
@@ -67,7 +76,9 @@ _zsh_highlight()
 
   # Remove all highlighting in isearch, so that only the underlining done by zsh itself remains.
   # For details see FAQ entry 'Why does syntax highlighting not work while searching history?'.
-  if [[ $WIDGET == zle-isearch-update ]] && ! (( $+ISEARCHMATCH_ACTIVE )); then
```
m0vie: We should probably add a comment here stating that we must never use any pattern before this line, in order to not break things again.

danielshahaf:
@docwhat The next zsh feature release will be 5.4.  There may be a 5.3.2 hotfix release before then.  In principle, zsh may decide to release a 5.3.1.1, but that's unlikely.  And in either of the three cases, we can't be sure whether the release would fix the `PAT_STATIC` bug: @m0vie's patch might be reverted or the bug accidentally reintroduced before upstream's release.

Now, I won't release z-sy-h with an is-at-least condition that refers to a _future_ zsh release, and people who use zsh from master should use HEAD of master; therefore, I think it doesn't matter what version number we pass to is-at-least, so long as we have a note on #412 to fix it after zsh's release and before z-sy-h's release.

Makes sense?

zsh-syntax-highlighting.zsh
danielshahaf: @m0vie At this point it may be easier to ask pws to release a 5.3.2 with your patch (and other stability fixes) included?

README.md
danielshahaf: (being discussed in the toplevel comments below)

docwhat:
@danielshahaf:

Just so I understand what my action item is: I should set the `is-at-least` check to `5.3.1.1` (as you suggested previously?

Can we push zsh to release 5.3.2 sooner rather than later and release it with that version check (as you mentioned above)?

danielshahaf:
> @danielshahaf:
> 
> Just so I understand what my action item is: I should set the
> `is-at-least` check to `5.3.1.1` (as you suggested previously?

My current idea is that it _doesn't matter_ whether the code says `is-at-least 5.3.1.1` or `is-at-least 5.3.2`, since I will wait for the next zsh release — and update the argument to is-at-least accordingly — before releasing z-sy-h.  Makes sense?

> Can we push zsh to release 5.3.2 sooner rather than later and release it with that version check (as you mentioned above)?

We certainly can.  All it takes is for someone to email zsh-workers@ and ask/convince them to do so. :)

docwhat:
It's been a while, @danielshahaf -- Is there anything else I need to do with this PR? Or are we still good?

danielshahaf:
@movie's patch is [workers/40453](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=40453), zsh-users/zsh@e51c9c17af51e4055efb5a2cc36739d1d7ae457f.

danielshahaf:
@docwhat Apologies for dropping the ball.  Merged this and updated #415 accordingly.

If anyone wants to push upstream for a 5.3.2 (as per my penultimate comment), feel free :-)

README.md
m0vie: The information in the readme is still wrong. See my comment above.

danielshahaf: Thanks. Broke this out as #423

docwhat:
> @docwhat Apologies for dropping the ball.

No apologies necessary! Thanks!

PR #417, opened 2017-02-18
fix file permissions

gnowxilef:
causes compinit errors


danielshahaf:
Thanks for the patch.  I guess you've figured it by now: `tests/` shouldn't be in $fpath.

PR #426, opened 2017-04-09
fix printing error message to file  when cannot resolve highlighters …

jsoizo:


danielshahaf:
Thanks @jsoizo, I've merged this with a tweaked log message.    Also thanks @phy1729 for reviewing/approving over IRC.

PR #441, opened 2017-09-08
Fix links in INSTALL.md

dukex:
Just fixing links in INSTALL.md


danielshahaf:
@dukex Good catch.  I'd rather use `[foo][bar]` style links, in order to keep INSTALL.md readable in source form.  It'd be great if you could do this, else, I'll make it happen when I have a minute.  Cheers!

PR #451, opened 2017-11-03
Fixes to be NOUNSET clean

phy1729:
There are still a few more changes required (on my nounset-full branch) to default CURSOR, PENDING, and WIDGET to empty; but they may be better fixed upstream in zsh.

### DIFF
--- a/tests/test-highlighting.zsh
+++ b/tests/test-highlighting.zsh
@@ -135,12 +135,16 @@ run_test_internal() {
     integer start=$highlight_zone[1] end=$highlight_zone[2]
     # Escape # as ♯ since the former is illegal in the 'description' part of TAP output
     local desc="[$start,$end] «${BUFFER[$start,$end]//'#'/♯}»"
-    # Match the emptiness of observed_result if no highlighting is expected
-    [[ $highlight_zone[3] == NONE ]] && highlight_zone[3]=
     (( $+highlight_zone[4] )) && todo="# TODO $highlight_zone[4]"
     for j in {$start..$end}; do
-      if [[ "$observed_result[$j]" != "$highlight_zone[3]" ]]; then
-        print -r -- "not ok $i - $desc - expected ${(qqq)highlight_zone[3]}, observed ${(qqq)observed_result[$j]}. $todo"
+      if
+	if [[ $highlight_zone[3] == NONE ]]; then
+	  (( $+observed_result[$j] ))
+	else
+	  [[ "$observed_result[$j]" != "$highlight_zone[3]" ]]
+	fi
+      then
+        print -r -- "not ok $i - $desc - expected ${(qqq)highlight_zone[3]}, observed ${(qqq)observed_result[$j]-NONE}. $todo"
         continue 2
       fi
     done
--- a/tests/test-highlighting.zsh
+++ b/tests/test-highlighting.zsh
@@ -103,6 +103,9 @@ run_test_internal() {
   # Check the data declares $expected_region_highlight.
   (( ${#expected_region_highlight} == 0 )) && { echo >&2 "Bail out! 'expected_region_highlight' is not declared or empty."; return 1; }
 
+  # Set sane defaults for ZLE variables
+  : ${CURSOR=$#BUFFER} ${PENDING=0} ${WIDGET=z-sy-h-test-harness-test-widget}
+
   # Process the data.
   region_highlight=()
   _zsh_highlight
--- a/tests/test-highlighting.zsh
+++ b/tests/test-highlighting.zsh
@@ -29,6 +29,8 @@
 # -------------------------------------------------------------------------------------------------
 
 
+setopt NOUNSET
+
 # Check an highlighter was given as argument.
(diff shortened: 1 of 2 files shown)


phy1729:
Since zsh only promises to set CURSOR, PENDING, and WIDGET (among others) while executing a zle widget, it is not a zsh bug. Rebased to add sane defaults for these variables that can be overridden by the test. Tests are now run with NOUNSET.

highlighters/brackets/brackets-highlighter.zsh
```diff
@@ -59,11 +59,15 @@ _zsh_highlight_highlighter_brackets_paint()
...
-          matching[$pos]=$matchingpos
+        if (( level > 0 )); then
+          matchingpos=$lastoflevel[$level]
+          levelpos[$pos]=$((level--))
+          if _zsh_highlight_brackets_match $matchingpos $pos; then
+            matching[$matchingpos]=$pos
+            matching[$pos]=$matchingpos
+          fi
+        else
+          levelpos[$pos]=-1
```
danielshahaf: Can you explain this -1?  Why couldn't this assignment be removed?  (And then I wouldn't have to start thinking whether the modulo later is going to do `(-1-1) % N + 1` and what works out to...)

zsh-syntax-highlighting.zsh
```diff
@@ -188,7 +188,7 @@ _zsh_highlight_apply_zle_highlight() {
   integer first="$3" second="$4"
 
   # read the relevant entry from zle_highlight
-  local region="${zle_highlight[(r)${entry}:*]}"
+  local region="${zle_highlight[(r)${entry}:*]-}"
```
danielshahaf: Independent/preexisting issue: needs to be ${(b)entry}.

highlighters/brackets/brackets-highlighter.zsh
phy1729: No real reason other than it's an invalid value. There needs to be an assignment so the error is highlighted (the loop is over `${(k)levelpos}`). The modulo branch won't be reached since `(( $+matching[$pos] ))` will always fail.

zsh-syntax-highlighting.zsh
phy1729: That should be it's own commit though.

highlighters/brackets/brackets-highlighter.zsh
danielshahaf: +1

tests/test-highlighting.zsh
```diff
@@ -137,10 +142,10 @@ run_test_internal() {
     local desc="[$start,$end] «${BUFFER[$start,$end]//'#'/♯}»"
     # Match the emptiness of observed_result if no highlighting is expected
     [[ $highlight_zone[3] == NONE ]] && highlight_zone[3]=
-    [[ -n "$highlight_zone[4]" ]] && todo="# TODO $highlight_zone[4]"
+    (( $+highlight_zone[4] )) && todo="# TODO $highlight_zone[4]"
     for j in {$start..$end}; do
-      if [[ "$observed_result[$j]" != "$highlight_zone[3]" ]]; then
-        print -r -- "not ok $i - $desc - expected ${(qqq)highlight_zone[3]}, observed ${(qqq)observed_result[$j]}. $todo"
+      if [[ "${observed_result[$j]-}" != "$highlight_zone[3]" ]]; then
```
danielshahaf: I wonder if instead of ${foo-} here, we should explicitly test for undefinedness and pair it with the magic NONE value:

```zsh
if if (( ${+observed_result[$j]} )); then [[ $highlight_zone[3] == NONE ]]; else [[ $highlight_zone[3] == ${observed_result[$j] ]]; fi; then
```

(Well, rather, something equivalent to this but _slightly_ more readable ☺...)

zsh-syntax-highlighting.zsh
danielshahaf: Agreed.

tests/test-highlighting.zsh
```diff
@@ -135,12 +140,14 @@ run_test_internal() {
...
-    [[ $highlight_zone[3] == NONE ]] && highlight_zone[3]=
     (( $+highlight_zone[4] )) && todo="# TODO $highlight_zone[4]"
     for j in {$start..$end}; do
-      if [[ "$observed_result[$j]" != "$highlight_zone[3]" ]]; then
-        print -r -- "not ok $i - $desc - expected ${(qqq)highlight_zone[3]}, observed ${(qqq)observed_result[$j]}. $todo"
+      if if [[ $highlight_zone[3] == NONE ]]; then
+	(( $+observed_result[$j] ))
+      else
+	[[ "$observed_result[$j]" != "$highlight_zone[3]" ]]
+      fi; then
```
danielshahaf: Reindent so the inner if's `fi` isn't aligned with the outer's `if`?  Suggestion:

```sh
if if foo; then
     lorem
   else
     ipsum
   fi
then
  42
fi
```

PR #452, opened 2017-11-05
Set warn_create_global for tests

phy1729:


danielshahaf:
We certainly shouldn't leak parameters into calling scopes so +1 on the concept.

Should we try to use zsh/param/private?

PR #453, opened 2017-11-06
Add ability to skip tests

### LINKED ISSUE #382: Tests fail on msys2
https://gist.github.com/trapd00r/5586ef14b9d108ac439c4cdde9f19909

`option-path_dirs` fails because `chmod +x` is a no-op.

`path-broken-symlink` fails because `ln -s` is an alias for `cp` (and the test deliberately creates a broken symlink).

The tests should be fixed to either PASS or SKIP.

(Reported by @trapd00r.)

phy1729:
Closes #382. Needs to be tested on msys2 to ensure it fixes the issue.

### DIFF
--- a/highlighters/main/test-data/option-path_dirs.zsh
+++ b/highlighters/main/test-data/option-path_dirs.zsh
@@ -27,15 +27,18 @@
 # vim: ft=zsh sw=2 ts=2 et
 # -------------------------------------------------------------------------------------------------
 
-## setup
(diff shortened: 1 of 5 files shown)


danielshahaf:
+1

Osse:
Tested on MSYS2 running on Windows 10. Works fine.

phy1729:
Thanks for testing.

PR #457, opened 2017-11-25
main: Add fallback for useroptions if zsh/parameter is absent

phy1729:


highlighters/main/main-highlighter.zsh
```diff
@@ -195,7 +195,13 @@ _zsh_highlight_main__stack_pop() {
 _zsh_highlight_highlighter_main_paint()
 {
   # Before we even 'emulate -L', save the user's options
-  local -A useroptions; useroptions=("${(@kv)options}")
+  local -A useroptions
+  if zmodload -e zsh/parameter; then
```
danielshahaf: Maybe `|| zmodload zsh/parameter`?

highlighters/main/main-highlighter.zsh
```diff
@@ -195,7 +195,13 @@ _zsh_highlight_main__stack_pop() {
...
+    useroptions=("${(@kv)options}")
+  else
+    useroptions=(${(s: :)${(f)"$(setopt -m \*; setopt)"}/%on/off} ${(s: :)${(f)"$(setopt)"}/%/ on})
```
danielshahaf: Setting _all_ options!?  I can't imagine how that might behave unexpectedly...

(For starters, when I do `setopt -m \*` interactively it dumps vcs_info lines before the next prompt, since it sets XTRACE.)

phy1729:
The issue with `unsetopt` (and the current patch) is that it can return the no* variant of an option. Is there a clean way to get the canonical name (all lowercase, without extra `no`s, without underscores) of all known options without zsh/parameter?

highlighters/main/main-highlighter.zsh
phy1729: `if zmodload -e zsh/parameter; then` is already used in the main highlighter. The driver attempts to load the module, so I don't see a reason to attempt again.

danielshahaf:
Pretty sure `setopt` also emits leading `no` when unsetting an on by default option. Try 'set -o' (I think that's the letter but can't check.)

danielshahaf:
@phy1729 devised

```zsh
() { local a b; a=(${(f)"$(emulate -R zsh; set -o)"}); b=(${${${(M)a:#*off}%% *}#no} ${${(M)a:#*on}%% *}); b=( ${(o)b} ); typeset -p b }
```

which gets all option names (except for compatibility aliases such as DOT_GLOB).  Synthesising $options from that remains an exercise for... well, probably me, but I do wish there were a reader.

phy1729:
We could require zsh/parameter for now. If any users without zsh/parameter come forward, the quick fix would be to have an array of options used in conditionals and just test those in a loop as a fallback.

danielshahaf:
Not sure what you mean by a loop, but yes, a quick fallback would be to hardcode into $useroptions those specific options that we care about.

danielshahaf:
I think we need to add a `if ! zmodload -e zsh/parameter; then error out; fi` to the 'main' highlighter now?

danielshahaf:
> Not sure what you mean by a loop

`for option in $b; do [[ -o $option ]]; user_options+=($option ${${=:-off on}[2-$?]}); done`

The option names to beware of are:

```zsh
echo ${(k)options[(I)no*]} 
notify nomatch
```

phy1729:
PR updated. Not pretty, but it should actually work this time.

highlighters/main/main-highlighter.zsh
```diff
@@ -196,7 +196,21 @@ _zsh_highlight_main__stack_pop() {
...
+  local -A useroptions
+  if zmodload -e zsh/parameter; then
+    useroptions=("${(@kv)options}")
+  else
+    local canonicaloptions onoff option rawoptions
+    rawoptions=(${(f)"$(emulate -R zsh; set -o)"})
+    canonicaloptions=(${${${(M)rawoptions:#*off}%% *}#no} ${${(M)rawoptions:#*on}%% *})
+    for option in $canonicaloptions; do
+      [[ -o $option ]]
+      # This variable cannot be eliminated due to a zsh bug with $? in array assignment.
```
danielshahaf: Suggest to mention "workers/42101" in this comment.

highlighters/main/main-highlighter.zsh
```diff
@@ -196,7 +196,21 @@ _zsh_highlight_main__stack_pop() {
 _zsh_highlight_highlighter_main_paint()
 {
   # Before we even 'emulate -L', save the user's options
-  local -A useroptions; useroptions=("${(@kv)options}")
+  local -A useroptions
+  if zmodload -e zsh/parameter; then
+    useroptions=("${(@kv)options}")
+  else
+    local canonicaloptions onoff option rawoptions
+    rawoptions=(${(f)"$(emulate -R zsh; set -o)"})
```
danielshahaf: Is -R the right thing here?

highlighters/main/main-highlighter.zsh
```diff
@@ -196,7 +196,21 @@ _zsh_highlight_main__stack_pop() {
 _zsh_highlight_highlighter_main_paint()
 {
   # Before we even 'emulate -L', save the user's options
-  local -A useroptions; useroptions=("${(@kv)options}")
+  local -A useroptions
+  if zmodload -e zsh/parameter; then
+    useroptions=("${(@kv)options}")
+  else
+    local canonicaloptions onoff option rawoptions
```
danielshahaf: Suggest to wrap the contents of the `else` block into an anonymous function so these variables don't persist for the rest of `_zsh_highlight_highlighter_main_paint`.  The anonymous function will set $useroptions which will have been declared by its caller.

phy1729: I'm not sure it's worth it. If there is an error, the message would start `(anon)` rather than the more helpful `_zsh_highlight_highlighter_main_paint`.

danielshahaf: This should be fixed in zsh itself.

A block isn't a scope, right? `{ local a; } echo $+a` so either an explicit `unset` or punt. Your call.

PR #463, opened 2017-12-16
driver: Run under emulate -L zsh and add zsyh_user_options

phy1729:


danielshahaf:
+1

danielshahaf:
Should we doc that option aliases aren't supported?

phy1729:
Indeed we should. Added a sentence to specify how options must be specified.

danielshahaf:
LGTM

PR #467, opened 2017-12-23
'main': Fix highlighting of «<->»-style numeric globs.

danielshahaf:


phy1729:
This highlights `print 0<->` as a redirection.

danielshahaf:
... instead of as a glob pattern for "the character `0` followed by a number".

danielshahaf:
35270f7c97b9 ?

highlighters/main/test-data/number_range-glob.zsh
```diff
@@ -0,0 +1,37 @@
...
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+# IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
+# OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# -------------------------------------------------------------------------------------------------
+# -*- mode: zsh; sh-indentation: 2; indent-tabs-mode: nil; sh-basic-offset: 2; -*-
+# vim: ft=zsh sw=2 ts=2 et
+# -------------------------------------------------------------------------------------------------
+
+BUFFER=$'print <-> 0<->'
```
phy1729: Perhaps add a `<foo->` which is two redirections.

phy1729:
Would incorrectly match `foo<bar-baz>quux`, but `(z)` saves us by splitting the redirects into separate words. Needs a rebase post arg-quoting merge.

highlighters/main/test-data/number_range-glob.zsh
danielshahaf: Added.

danielshahaf:
> Would incorrectly match foo<bar-baz>quux, but (z) saves us by splitting the redirects into separate words.

So, no changes needed, right?

> Needs a rebase post arg-quoting merge.

Yeah, and a squash.  Will do.

phy1729:
> So, no changes needed, right?
Yep. Just the rebase.

danielshahaf:
The rebase is blocked by #468.

danielshahaf:
The rebase is not trivial, the logic moved to a characterwise loop in `_zsh_highlight_main_highlighter_highlight_argument`.

That doesn't get called on redirections (https://github.com/zsh-users/zsh-syntax-highlighting/blob/df0f44858634c454fbbabc182fc5ffe43b8907b5/highlighters/main/main-highlighter.zsh#L695-L698) so any `<` not quoted (by backslash or quotes) should be part of the `<->` globbing syntax.

danielshahaf:
Okay, how about https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:numeric-globs-v2 (0decb2a8f541)?

danielshahaf:
Merged after @phy1729 review.  (The PR is "closed"/red rather than "merged"/purple because it's about -v1 and it's -v2 that got merged.)

danielshahaf:
The commit is 987b743646c3e71a160de0ac2177bfe914bde39a in master.

PR #470, opened 2017-12-25
driver: Highlight even after a widget returns non-zero

### LINKED ISSUE #90: Unsuccessful completion removes highlighting
If completion returns an error all highlighting will be removed. 
Eg. failed file expansion `cat filethatdoesntexist<tab>`

This patch fixes it:

```
index cbd63e9..4ad0106 100644
@@ -105,9 +105,12 @@ typeset -gA ZSH_HIGHLIGHT_STYLES
 # Whether the command line buffer has been modified or not.
 #
 # Returns 0 if the buffer has changed since _zsh_highlight was last called.
+# This doesn't work reliably. As the buffer might be modified, but without
+# changing the text content. Eg. when file expansion fail when
+# autocompelting.
 _zsh_highlight_buffer_modified()
 {
-  [[ "${_ZSH_HIGHLIGHT_PRIOR_BUFFER:-}" != "$BUFFER" ]]
+  true
 }

 # Whether the cursor has moved or not.
@@ -147,7 +150,7 @@ _zsh_highlight_bind_widgets()

       # Completion widget: override and rebind old one with prefix "orig-".
       completion:*) eval "zle -C orig-$cur_widget ${${widgets[$cur_widget]#*:}/:/ }; \
-                          _zsh_highlight_widget_$cur_widget() { builtin zle orig-$cur_widget -- \"\$@\" && _zsh_highlight }; \
+                          _zsh_highlight_widget_$cur_widget() { { builtin zle orig-$cur_widget -- \"\$@\"; } always { _zsh_highlight; } }; \
                           zle -N $cur_widget _zsh_highlight_widget_$cur_widget";;

       # Builtin widget: override and make it call the builtin ".widget".
```

phy1729:
This needs some thought and discussion before (if) merging. May have yet unconsidered unintended side effects.

Closes #90.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -286,8 +286,10 @@ _zsh_highlight_add_highlight()
 # $1 is name of widget to call
 _zsh_highlight_call_widget()
 {
-  builtin zle "$@" && 
+  builtin zle "$@"
+  local -r _zsyh_wrappee_status=$?
   _zsh_highlight
+  return _zsyh_wrappee_status
 }
 
 # Rebind all ZLE widgets to make them invoke _zsh_highlights.


danielshahaf:
As discussed on IRC:

The `&&` goes back to the very first commit by @nicoulaj: 89e3c632ef550f33534a935515519f8d14aa1fef line 87

I guess the rationale is that after a general widget returns an error, "Do no harm" kicks in so we avoid setting $region_highlight since we don't know that zle's in a defined state.

Can we distinguish "Completion had no matches" from "Completion had an error"?  E.g., for completion widgets, check ${compstate[nmatches]}, or _something_...  (maybe ask zsh-users@)

Thinking out loud — how about if we restrict this to completion widgets?  E.g., those that appear in `zle -lL` or `${(k)widgets[(R)completion:*]}`?

danielshahaf:
The incumbent behaviour is actually useful, in a way: typing `rsync -4 -6 --<TAB>` (which is an error: -4 and -6 are declared mutually exclusive in `_rsync`) currently gives no indication of the error other than the command word changing from green to uncoloured.

Sure, in a way that's an upstream bug (unclear error reporting), but leaving the command word green might nevertheless be a usability regression.

/cc @blueyed (context: discussion on #zsh today)

danielshahaf:
If we change this it would be good to expose the wrappee's code to highlighters, e.g., —

```diff
 _zsh_highlight_call_widget()
 {
-  builtin zle "$@" && 
+  builtin zle "$@"
+  local zsh_highlight_wrappee_status=$?
   _zsh_highlight
+  return zsh_highlight_wrappee_status
 }
```

— although actually reporting the error condition isn't our responsibility (it's the responsibility of the wrappee, and it doesn't communicate the error to the user through $region_highlight so we don't step on its toes).

phy1729:
Asked upstream in [users/23100](http://zsh.org/mla/users/2018/msg00059.html).

danielshahaf:
zsh master does print a warning nowadays, doesn't it? Which would address the above 'usability regression' point.

phy1729:
There was a patch to `_arguments` and a suggestion to modify `_main_complete` instead, but as far as I can tell no changes made their way to git.

Issue #365, opened 2016-09-15
Treat doas just like sudo


hsiktas:
On some BSDs sudo is not included by default and now with [doas](http://man.openbsd.org/doas.1) we have a similar alternative which is somewhat easier to configure.
Can we treat doas just like sudo for syntax highlighting?

danielshahaf:
Just pushed a fix to an outstanding feature branch.

Issue #343, opened 2016-07-28
flags to `command` are mishighlighted as missing commands


anntzer:
Basically,

```
$ command -v foo
```

highlights `-v` in red (as I don't have an executable called "-v" in my PATH).

danielshahaf:
Similar to #168 #247 #248.  Fixing this in the current infrastructure will require extending the state machine; but extending the state machine for each of _N_ precommands isn't the best idea.  Perhaps there's a better option (something based on `getopts` or `zparseopts` might be nice).

The other requests _could_ be addressed if they are normally invoked without options (so `nice pwd`, `doas pwd`, `pkexec pwd` would work but `foo -x pwd` would highlight the `-x` instead of the `pwd` as a command).  Perhaps that should be done...

anntzer:
Sorry, I skimmed through the tracker for a duplicate but couldn't find one.  Feel free to close this.

danielshahaf:
I've added a test (0d41933) and started on a fix (https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:i343-precommands-v1), but ran into an outstanding issue about aliases not being resolved correctly.

danielshahaf:
That outstanding issue is #264.

danielshahaf:
I've force-pushed that branch, it now fixes both `command` and `nice`, and adds support for aliases too.  The aliases handling needs a closer look, (a) audit uses of `$arg`, (b) outstanding questions on #247.

danielshahaf:
That branch now conflicts due to:

* 2c002f9 noop: Add comments.
* 757d047 'main': The word after 'sudo' is only a non-command word if it is an option.
* 179b8e7 'main': Test for redirection earlier.
* d1c773f noop: Fix indentation.

danielshahaf:
I"ve rebased the branch as https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:i343-precommands-v2.

It's a pretty invasive change so I'm tentatively scheduling it for 0.6.0, but with favourable winds and calm seas it could get merged in time for 0.5.0.

ascii-soup:
I've been testing this fix against master side-by-side and I couldn't find any instances where the new behaviour was incorrect (as far as this issue goes). Overall it works better / is fixed versus master.
:+1:

danielshahaf:
[...]

Issue #264, opened 2016-01-11
command word: Order of expansions is wrong


danielshahaf:
https://github.com/zsh-users/zsh-syntax-highlighting/blob/5e4b446dee770a811ffb6c19ee41f624895e56da/highlighters/main/main-highlighter.zsh#L275-L300

As [zshexpn(1)](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Expansion) specifies, alias expansion precedes parameter expansion which precedes tilde expansion.  However, the code implements them in a different order (e.g., tilde expansion is performed before alias expansion).

phy1729:
The expansion order is now: alias, tilde, parameter; however one should note that
```zsh
foo=~root
echo $foo
```
does not expand to `/root`, so we should keep track of quoting if the order is fixed.

danielshahaf:
> The expansion order is now: alias, tilde, parameter; however one should 
> note that
> ```zsh
> foo=~root
> echo $foo
> ```
> does not expand to `/root`, so we should keep track of quoting if the 
> order is fixed.

How do you reproduce that?  I tried { assignment in command word position, `local` the builtin, and `local` the keyword } x { magicequalsubst set, magicequalsubst unset } and in all of these cases `typeset -p foo` showed "/root".

phy1729:
I updated the comment shortly after posting to be `foo='~root'` because it was wrong without the quotes.

danielshahaf:
I think nowadays we could implement that easily by having `_zsh_highlight_main_highlighter_expand_path` check `(( in_param ))`.  Test case: set `a=/usr` then type `$a` in command position (without a trailing space).

danielshahaf:
Triage: https://github.com/zsh-users/zsh-syntax-highlighting/pull/471#issuecomment-632473281

Issue #168, opened 2015-04-27
highlight `nice foo`


anntzer:
Similarly to #107, which highlights the command after `sudo`, one could highlight the command after `nice`.

danielshahaf:
The problem is options parsing: see 19981ef9eae354ee004cd6b9a54bf07c0a94965c and 3b3a58be9114687d21f03840ebc4c53365d5e604 for an analogous example.

There are many other commands that accept an `execve`-able argument list; some common ones are `env`, `xargs`, and `find ... -exec`.  Perhaps it would be better, instead of hardcoding _N_ separate option parsers, to hook into completion system (issue #10) and ask compsys to identify command words for us?

compsys already solved the "parse `find`'s command line" problem, and I don't want to have to repeat that exercise. ☺

[ That said... if it's possible to support _nice_, or _env_, or any other single utility, in a simple manner without adding lots of twisty codepaths, that might be worthwhile as an interim measure until #10 is fixed ]

anntzer:
Let's add `strace` to the list of affected commands, too.

PR #471, opened 2017-12-25
Highlight precommands and correct expansions order

danielshahaf:
This is i343-precommands-v2, but rebased, and with an extra commit about ${foo} parameter expansions.

Related to #264, #328, #343, #168, #365 (see log messages and added comments).

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -214,7 +214,7 @@ _zsh_highlight_highlighter_main_paint()
   fi
 
   ## Variable declarations and initializations
-  local start_pos=0 end_pos highlight_glob=true arg style
+  local start_pos=0 end_pos highlight_glob=true arg arg_raw style
   local in_array_assignment=false # true between 'a=(' and the matching ')'
   typeset -a ZSH_HIGHLIGHT_TOKENS_COMMANDSEPARATOR
   typeset -a ZSH_HIGHLIGHT_TOKENS_PRECOMMANDS
@@ -322,6 +322,9 @@ _zsh_highlight_highlighter_main_paint()
     args=(${(z)buf})
   fi
   for arg in $args; do
+    # Save an unmunged copy of the current word.
+    arg_raw="$arg"
+
     # Initialize $next_word.
     if (( in_redirection )); then
       (( --in_redirection ))
@@ -432,6 +435,30 @@ _zsh_highlight_highlighter_main_paint()
       fi
     fi
 
+    # Expand aliases.
+    # TODO: this should be done iteratively, e.g., 'alias x=y y=z z=w\n x' 
+    #       And then the entire 'alias' branch of the 'case' statement should
+    #       be done here.
+    () {
+      # TODO: path expansion should happen _after_ alias expansion
+      _zsh_highlight_main_highlighter_expand_path $arg
+      local expanded_arg="$REPLY"
+      _zsh_highlight_main__type ${expanded_arg}
+    }
+    local res="$REPLY"
+    if [[ $res == "alias" ]]; then
+      _zsh_highlight_main__resolve_alias $arg
+      () {
+        # Use a temporary array to ensure the subscript is interpreted as
+        # an array subscript, not as a scalar subscript
+        local -a reply
+        # TODO: the ${interactive_comments+set} path needs to skip comments; see test-data/alias-comment1.zsh
+        reply=( ${interactive_comments-${(z)REPLY}}
+                ${interactive_comments+${(zZ+c+)REPLY}} )
+        arg=$reply[1]
+      }
+    fi
+
     # Special-case the first word after 'sudo'.
     if (( ! in_redirection )); then
       if [[ $this_word == *':sudo_opt:'* ]] && [[ $arg != -* ]]; then
@@ -472,10 +499,6 @@ _zsh_highlight_highlighter_main_paint()
       next_word+=':sudo_opt:'
       next_word+=':start:'
      else
-      _zsh_highlight_main_highlighter_expand_path $arg
-      local expanded_arg="$REPLY"
-      _zsh_highlight_main__type ${expanded_arg}
-      local res="$REPLY"
       () {
         # Special-case: command word is '$foo', like that, without braces or anything.
         #
@@ -548,8 +571,9 @@ _zsh_highlight_highlighter_main_paint()
                         ;;
         'suffix alias') style=suffix-alias;;
         alias)          () {
+                          # Make sure to use $arg_raw here, rather than $arg.
                           integer insane_alias
-                          case $arg in
+                          case $arg_raw in
                             # Issue #263: aliases with '=' on their LHS.
                             #
                             # There are three cases:
@@ -563,12 +587,13 @@ _zsh_highlight_highlighter_main_paint()
                           esac
                           if (( insane_alias )); then
                             style=unknown-token
+                          # Calling 'type' again; since __type memoizes the answer, this call is just a hash lookup.
+                          elif _zsh_highlight_main__type "$arg"; [[ $REPLY == 'none' ]]; then
+                            style=unknown-token
                           else
                             # The common case.
                             style=alias
-                            _zsh_highlight_main__resolve_alias $arg
-                            local alias_target="$REPLY"
-                            [[ -n ${(M)ZSH_HIGHLIGHT_TOKENS_PRECOMMANDS:#"$alias_target"} && -z ${(M)ZSH_HIGHLIGHT_TOKENS_PRECOMMANDS:#"$arg"} ]] && ZSH_HIGHLIGHT_TOKENS_PRECOMMANDS+=($arg)
+                            [[ -n ${(M)ZSH_HIGHLIGHT_TOKENS_PRECOMMANDS:#"$arg"} && -z ${(M)ZSH_HIGHLIGHT_TOKENS_PRECOMMANDS:#"$arg_raw"} ]] && ZSH_HIGHLIGHT_TOKENS_PRECOMMANDS+=($arg_raw)
                           fi
                         }
                         ;;
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -243,6 +243,7 @@ _zsh_highlight_highlighter_main_paint()
     'command' ''
     'nice' n
     'sudo' Cgprtu
+    'doas' aCu
   )
 
   if [[ $zsyh_user_options[ignorebraces] == on || ${zsyh_user_options[ignoreclosebraces]:-off} == on ]]; then
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -289,8 +289,9 @@ _zsh_highlight_highlighter_main_paint()
   #
   # The states are:
   # - :start:      Command word
-  # - :sudo_opt:   A leading-dash option to sudo (such as "-u" or "-i")
-  # - :sudo_arg:   The argument to a sudo leading-dash option that takes one,
+  # - :sudo_opt:   A leading-dash option to a precommand, whether it takes an
+  #                argument or not.  (Example: sudo's "-u" or "-i".)
+  # - :sudo_arg:   The argument to a precommand's leading-dash option,
   #                when given as a separate word; i.e., "foo" in "-u foo" (two
   #                words) but not in "-ufoo" (one word).
   # - :regular:    "Not a command word", and command delimiters are permitted.
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -469,8 +469,14 @@ _zsh_highlight_highlighter_main_paint()
       # parameters that refer to commands, functions, and builtins.
       local -a match mbegin mend
       local MATCH; integer MBEGIN MEND
+      local parameter_name
+      if [[ $arg[1] == '$' ]] && [[ ${arg[2]} == '{' ]] && [[ ${arg[-1]} == '}' ]]; then
+        parameter_name=${${arg:2}%?}
+      elif [[ $arg[1] == '$' ]]; then
+        parameter_name=${arg:1}
+      fi
       if [[ $res == none ]] && (( ${+parameters} )) &&
-         [[ ${arg[1]} == \$ ]] && [[ ${arg:1} =~ ^([A-Za-z_][A-Za-z0-9_]*|[0-9]+)$ ]] &&
+         [[ ${parameter_name} =~ ^([A-Za-z_][A-Za-z0-9_]*|[0-9]+)$ ]] &&
          (( ${+parameters[${MATCH}]} ))
          then
         arg=${(P)MATCH}
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -505,17 +505,20 @@ _zsh_highlight_highlighter_main_paint()
     # Parse the sudo command line
     if (( ! in_redirection )); then
       if [[ $this_word == *':sudo_opt:'* ]]; then
-        case "$arg" in
+        if [[ -n $flags_with_argument ]] &&
+           [[ $arg == '-'[$flags_with_argument] ]]; then
           # Flag that requires an argument
-          '-'[$flags_with_argument])
-                       this_word=${this_word//:start:/};
-                       next_word=':sudo_arg:';;
+          this_word=${this_word//:start:/}
+          next_word=':sudo_arg:'
+        elif [[ $arg == '-'* ]]; then
+          # Flag that requires no argument, or unknown flag.
           # This prevents misbehavior with sudo -u -otherargument
-          '-'*)        this_word=${this_word//:start:/};
-                       next_word+=':start:';
-                       next_word+=':sudo_opt:';;
-          *)           ;;
-        esac
+          this_word=${this_word//:start:/}
+          next_word+=':start:'
+          next_word+=':sudo_opt:'
+        else
+          #
+        fi
       elif [[ $this_word == *':sudo_arg:'* ]]; then
         next_word+=':sudo_opt:'
         next_word+=':start:'
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -241,6 +241,7 @@ _zsh_highlight_highlighter_main_paint()
   local -A precommand_options
   precommand_options=(
     'command' ''
+    'nice' n
     'sudo' Cgprtu
   )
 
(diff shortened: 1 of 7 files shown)


danielshahaf:
Tests pass locally.

highlighters/main/main-highlighter.zsh
```diff
@@ -432,6 +449,56 @@ _zsh_highlight_highlighter_main_paint()
       fi
     fi
 
+    # Expand aliases.
+    # TODO: this should be done iteratively, e.g., 'alias x=y y=z z=w\n x'
```
phy1729: Trailing whitespace here.

highlighters/main/main-highlighter.zsh
```diff
@@ -432,6 +449,56 @@ _zsh_highlight_highlighter_main_paint()
...
     fi
 
+    # Expand aliases.
+    # TODO: this should be done iteratively, e.g., 'alias x=y y=z z=w\n x' 
+    #       And then the entire 'alias' branch of the 'case' statement should
+    #       be done here.
+    () {
+      # TODO: path expansion should happen _after_ alias expansion
+      _zsh_highlight_main_highlighter_expand_path $arg
+      local expanded_arg="$REPLY"
```
phy1729: Any reason to keep this variable?

highlighters/main/main-highlighter.zsh
```diff
@@ -432,6 +449,56 @@ _zsh_highlight_highlighter_main_paint()
...
+        # TODO: the ${interactive_comments+set} path needs to skip comments; see test-data/alias-comment1.zsh
+        reply=( ${interactive_comments-${(z)REPLY}}
+                ${interactive_comments+${(zZ+c+)REPLY}} )
+        arg=$reply[1]
+      }
+    fi
+
+    # Expand parameters.
+    #
+    # ### For now, expand just '$foo', like that, without braces or anything.
```
phy1729: Comment was incorrect prior to the move as well (`${foo}` is handled, but nothing more complicated).

highlighters/main/main-highlighter.zsh
```diff
@@ -432,6 +449,56 @@ _zsh_highlight_highlighter_main_paint()
...
+      if [[ $arg[1] == '$' ]] && [[ ${arg[2]} == '{' ]] && [[ ${arg[-1]} == '}' ]]; then
+        parameter_name=${${arg:2}%?}
+      elif [[ $arg[1] == '$' ]]; then
+        parameter_name=${arg:1}
+      fi
+      if [[ $res == none ]] && (( ${+parameters} )) &&
+         [[ ${parameter_name} =~ ^([A-Za-z_][A-Za-z0-9_]*|[0-9]+)$ ]] &&
+         (( ${+parameters[${MATCH}]} ))
+         then
+        arg=${(P)MATCH}
```
phy1729: This will mangle arrays (can be handled in another commit since this is a prior issue).

highlighters/main/main-highlighter.zsh
```diff
@@ -442,16 +509,20 @@ _zsh_highlight_highlighter_main_paint()
     # Parse the sudo command line
     if (( ! in_redirection )); then
       if [[ $this_word == *':sudo_opt:'* ]]; then
-        case "$arg" in
+        if [[ -n $flags_with_argument ]] &&
+           [[ $arg == '-'[$flags_with_argument] ]]; then
           # Flag that requires an argument
-          '-'[Cgprtu]) this_word=${this_word//:start:/};
-                       next_word=':sudo_arg:';;
+          this_word=${this_word//:start:/}
```
phy1729: Any reason to not use `this_word=':sudo_arg:'` (here and below)?

highlighters/main/main-highlighter.zsh
```diff
@@ -442,16 +509,20 @@ _zsh_highlight_highlighter_main_paint()
     # Parse the sudo command line
     if (( ! in_redirection )); then
       if [[ $this_word == *':sudo_opt:'* ]]; then
-        case "$arg" in
+        if [[ -n $flags_with_argument ]] &&
+           [[ $arg == '-'[$flags_with_argument] ]]; then
```
phy1729: This will not correctly highlight `doas -nu phy1729 echo`. Not a regression.

highlighters/main/main-highlighter.zsh
```diff
@@ -442,16 +509,20 @@ _zsh_highlight_highlighter_main_paint()
...
+        elif [[ $arg == '-'* ]]; then
+          # Flag that requires no argument, or unknown flag.
           # This prevents misbehavior with sudo -u -otherargument
-          '-'*)        this_word=${this_word//:start:/};
-                       next_word+=':start:';
-                       next_word+=':sudo_opt:';;
-          *)           ;;
-        esac
+          this_word=${this_word//:start:/}
+          next_word+=':start:'
```
phy1729: Might as well combine the two appends.

highlighters/main/main-highlighter.zsh
```diff
@@ -442,16 +509,20 @@ _zsh_highlight_highlighter_main_paint()
...
+          next_word+=':sudo_opt:'
+        else
```
phy1729: Any reason for the empty else?

highlighters/main/main-highlighter.zsh
```diff
@@ -464,35 +535,15 @@ _zsh_highlight_highlighter_main_paint()
      style=reserved-word # de facto a reserved word, although not de jure
      next_word=':start:'
    elif [[ $this_word == *':start:'* ]] && (( in_redirection == 0 )); then # $arg is the command word
-     if [[ -n ${(M)ZSH_HIGHLIGHT_TOKENS_PRECOMMANDS:#"$arg"} ]]; then
-      style=precommand
-     elif [[ "$arg" = "sudo" ]] && { _zsh_highlight_main__type sudo; [[ -n $REPLY && $REPLY != "none" ]] }; then
+     if (( ${+precommand_options[$arg]} )) && { _zsh_highlight_main__type $arg; [[ -n $REPLY && $REPLY != "none" ]] }; then
```
phy1729: `{ _zsh_highlight_main__type $arg; [[ -n $REPLY && $REPLY != "none" ]] }` is non obvious in purpose. Perhaps a comment should be added (or have __type return 1 if the command doesn't exist). Prior issue; doesn't block merging.

highlighters/main/main-highlighter.zsh
```diff
@@ -464,35 +535,15 @@ _zsh_highlight_highlighter_main_paint()
...
       style=precommand
+      flags_with_argument=${precommand_options[$arg]}
       next_word=${next_word//:regular:/}
       next_word+=':sudo_opt:'
       next_word+=':start:'
+     elif [[ -n ${(M)ZSH_HIGHLIGHT_TOKENS_PRECOMMANDS:#"$arg"} ]]; then
```
phy1729: Could we instead combine `ZSH_HIGHLIGHT_TOKENS_PRECOMMANDS` into `precommand_options`?

phy1729:
Haven't had a chance to figure out why but when on this branch
```
% zsh -f
% . ./zsh-syntax-highlighting/zsh-syntax-highlighting.zsh 
% alias a=b
% : 'azsh-syntax-highlighting: BUG: _zsh_highlight_main_highlighter_expand_path: called without argument
'a'
```
That last line is typing `: 'a'`.

highlighters/main/main-highlighter.zsh
danielshahaf: I find `if {f; [[ -n $REPLY ]]}; then` idiomatic, especially with the braces. What would a comment say?

s/;/&&/ etc sounds good.

danielshahaf:
Note to self: audit for null argument elision.

highlighters/main/main-highlighter.zsh
danielshahaf: Also pasted arguments `doas -uphy1729`, right?

phy1729: `doas -uphy1729 echo` is highlights `echo` correctly since `-uphy1729` is not matched as a flag with argument.

highlighters/main/main-highlighter.zsh
phy1729: It took me a few seconds to realize that the point of the code was checking for the command existing on this system is all. Not attached to adding a comment if you don't think it necessary,

danielshahaf: Good idea.

highlighters/main/main-highlighter.zsh
danielshahaf: Mind if we fix this on master post merge? Along with the other ~10 instances in master?

highlighters/main/main-highlighter.zsh
danielshahaf: The argument would be readability through naming the return value, but I've removed it.

highlighters/main/main-highlighter.zsh
danielshahaf: Fixed.

highlighters/main/main-highlighter.zsh
danielshahaf: Right; that's actually tested: https://github.com/danielshahaf/zsh-syntax-highlighting/blob/d838a29df342e9c02f2d006e63a5b1eb5210cfd7/highlighters/main/test-data/precommand3.zsh#L30

danielshahaf: The problem with `doas -nu phy1729 echo` is that we can't currently distinguish a `-n` flag that takes no argument from a `-n` flag that takes a `u` argument followed by `phy1729 echo` as two positional arguments, since we only know which flag-letters take arguments.

We could do something like... *coding noises* 2c15b0e996ad.

highlighters/main/main-highlighter.zsh
danielshahaf: That's just coding style, isn't it?

highlighters/main/main-highlighter.zsh
danielshahaf: How about now?

highlighters/main/main-highlighter.zsh
danielshahaf: Done

highlighters/main/main-highlighter.zsh
danielshahaf: 1bf54d7f268b

highlighters/main/main-highlighter.zsh
danielshahaf: Where's below?  Did you mean `this_word=':sudo_opt:'`?

highlighters/main/main-highlighter.zsh
danielshahaf: e9ab802593eb — passes all tests but please review.

danielshahaf:
The problem is here:

https://github.com/danielshahaf/zsh-syntax-highlighting/blob/e9ab802593ebbfce5cf749b1854d2cc201b3aa68/highlighters/main/main-highlighter.zsh#L471-L482

$arg starts out as `'a'` (three characters), then `_zsh_highlight_main_highlighter_expand_path` dequotes, `_zsh_highlight_main__type` returns `alias`, and `_zsh_highlight_main__resolve_alias` doesn't find an alias named `'a'` (there's only an `a` alias) and returns an empty string, which is elided in `_zsh_highlight_main_highlighter_check_path` at https://github.com/danielshahaf/zsh-syntax-highlighting/blob/e9ab802593ebbfce5cf749b1854d2cc201b3aa68/highlighters/main/main-highlighter.zsh#L850-L852 .

The breakage is that alias lookup should be done on the non-dequoted argument.  Not sure how to fix (ideas?).

Good catch!

danielshahaf:
Filed #479 which would affect the failure mode of this.

danielshahaf:
Remaining tasks:

- [ ] Resolve all review comments
- [ ] Update changelog
- [ ] Fix trailing whitespace
- [ ] `alias a=b; : 'a'` issue above, compare #479
- [ ] Tests fail under zsh<5.1
        # alias-comment1
        1..1
        not ok 1 - [1,1] «x» - expected "alias", observed "unknown-token". # TODO 'interactivecomments applies to aliases'

danielshahaf:
> Tests fail under zsh<5.1

Testing interactively in zsh-4.3.17, I get the expected output from running `x`, namely, `pwd` gets executed. So the test's expectations are correct for those old versions too. I don't immediately see the cause for the failure under 5.0.8.

highlighters/main/main-highlighter.zsh
phy1729: Works for me.

highlighters/main/main-highlighter.zsh
phy1729: :+1:

highlighters/main/main-highlighter.zsh
```diff
@@ -432,6 +468,65 @@ _zsh_highlight_highlighter_main_paint()
...
+        parameter_name=${arg:1}
+      fi
+      if [[ $res == none ]] && zmodload -e zsh/parameter &&
+         [[ ${parameter_name} =~ ^([A-Za-z_][A-Za-z0-9_]*|[0-9]+)$ ]] &&
+         (( ${+parameters[${MATCH}]} ))
+         then
+        # Set $arg.
+        case ${(tP)MATCH} in
+          (*array*|*assoc*)
+            local -a words=( ${(P)MATCH} )
```
phy1729: This will error on pre-5.1.

highlighters/main/main-highlighter.zsh
phy1729: Yes (and would be pedantically incorrect cf. next comment).

highlighters/main/main-highlighter.zsh
```diff
@@ -442,16 +537,23 @@ _zsh_highlight_highlighter_main_paint()
...
+          # Flag that requires no argument, or unknown flag.
           # This prevents misbehavior with sudo -u -otherargument
-          '-'*)        this_word=${this_word//:start:/};
-                       next_word+=':start:';
-                       next_word+=':sudo_opt:';;
-          *)           ;;
-        esac
+          this_word=${this_word//:start:/}
+          next_word+=':start:'
+          next_word+=':sudo_opt:'
```
phy1729: Perhaps to be pedantically correct ```(( $#precommand_options[$arg] )) && this_word+=':sudo_opt:'```.

highlighters/main/test-data/alias-comment1.zsh
```diff
@@ -0,0 +1,37 @@
+# -------------------------------------------------------------------------------------------------
+# Copyright (c) 2016 zsh-syntax-highlighting contributors
```
phy1729: 2018 (et seq).

highlighters/main/main-highlighter.zsh
```diff
@@ -563,12 +645,15 @@ _zsh_highlight_highlighter_main_paint()
                           esac
                           if (( insane_alias )); then
                             style=unknown-token
+                          # Calling 'type' again; since __type memoizes the answer, this call is just a hash lookup.
+                          elif _zsh_highlight_main__type "$arg" && [[ $REPLY == 'none' ]]; then
```
phy1729: Should this also be __is_runnable?

highlighters/main/test-data/alias-comment1.zsh
danielshahaf: No. Copyright year doesn't get incremented unless a copyrightable change is made. Rebase doesn't count.

danielshahaf: main highlighter does need a bump.

highlighters/main/main-highlighter.zsh
danielshahaf: Could, and check for $? = 1. Not sure whether it'd be clearer.

highlighters/main/main-highlighter.zsh
danielshahaf: What about `nice -x ls` where -x is an argumentless option not known to `precommand_options`?

danielshahaf:
Need precommand_options[nice]+=':'

(Can't add it as a review comment)

highlighters/main/main-highlighter.zsh
danielshahaf: Fixed, thanks.

(I spotted this late yesterday but ran into PEBKAGH trying to record it here)

highlighters/main/test-data/alias-comment1.zsh
danielshahaf: Done

highlighters/main/main-highlighter.zsh
phy1729: I don't know if we should assume all unknown options are argumentless. I think it'd be better to fail hard (-x is likely not an executable and so is unknown-token), so someone can either report it upstream or locally extend their `precommand_options`

highlighters/main/test-data/alias-comment1.zsh
phy1729: Ah, didn't realize this branch has been around that long.

highlighters/main/main-highlighter.zsh
phy1729: No need for the explicit `$? = 1`, `elif _zsh_highlight_main__is_runnable; then` is sufficient.

danielshahaf: [summary of IRC]

What should this do when `_zsh_highlight_main__type` fails (returns `$? = 1`)? Currently it falls back to the `alias` block below. Should it return `unknown-token`? Or maybe some new `fallback` style (since this isn't a syntactically-invalid BUFFER;, it's a "we don't know whether the command line is syntactically valid" situation)? In the latter case, should `$style` be initialised to `fallback` rather than `unknown-token`? (That would ripple to the opening condition of the non-command-word branch.)

phy1729: I think for now it'd be acceptable to `s/&&/||/` here and then open a new issue on `fallback`/`unknown-token`.

highlighters/main/main-highlighter.zsh
danielshahaf: This is appealing: it means the first unknown -x will be highlighted as an error regardless of whether or not it takes an argument.

I'll make it so.

danielshahaf:
Left:

- Resolve conflicts
- Address https://github.com/zsh-users/zsh-syntax-highlighting/pull/471#discussion_r161374006
- Merge
- Address https://github.com/zsh-users/zsh-syntax-highlighting/pull/471#discussion_r161105273

highlighters/main/main-highlighter.zsh
phy1729: Made so in e27e53b.

highlighters/main/main-highlighter.zsh
danielshahaf: This works in master and is covered by test-data/precommand4.zsh.

danielshahaf:
Triage: Incomplete; deferred ☹

PR #473, opened 2017-12-25
main: Highlight path after globbing checks

### LINKED ISSUE #468: path wrongly precedes globbing
```zsh
% touch \*
% echo *
```

The `*` is highlighted as `path`, it should be highlighted as `globbing`.

Bisected to dc1b2f6fa4bb477e7b9712ebdb4e82beb50258c6.

phy1729:
Closes #468.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -695,19 +695,14 @@ _zsh_highlight_highlighter_main_paint()
                  elif (( in_redirection == 2 )); then
                    style=redirection
                  else
-                   if _zsh_highlight_main_highlighter_check_path; then
-                     style=$REPLY
-                   else
-                     _zsh_highlight_main_highlighter_highlight_argument
-                     already_added=1
-                   fi
+                   _zsh_highlight_main_highlighter_highlight_argument
+                   already_added=1
                  fi
                  ;;
       esac
     fi
     if ! (( already_added )); then
       _zsh_highlight_main_add_region_highlight $start_pos $end_pos $style
-      [[ $style == path || $style == path_prefix ]] && _zsh_highlight_main_highlighter_highlight_path_separators
     fi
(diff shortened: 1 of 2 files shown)


danielshahaf:
This doesn't fix #468: https://github.com/zsh-users/zsh-syntax-highlighting/commit/ff61a496b5ea5f17eb0621314edb09219d421e44 still fails.  Interactively I see that it's now fg=blue,underline, compared to fg=default,underline in current master and fg=blue (no underline) before dc1b2f6fa4bb477e7b9712ebdb4e82beb50258c6.

Do we agree that _this specific case_ should be highlighted as it had been prior to dc1b2f6fa4bb477e7b9712ebdb4e82beb50258c6?

phy1729:
Ah I didn't know you had added a test and I failed to see the blue underline on a black background when testing interactively before. Fixed now.

danielshahaf:
Thanks. The new code looks good. Could you add a regression test for `noglob echo *` being highlighted as a path?

Also, `'f'oo` (which was discussed on #468) is now actually highlighted as `'f'` yellow+underline and `oo` default color + underline.  Would it be possible for you to add a test for that as well?  The zle internal combining logic might change in future zsh releases.

danielshahaf:
For posterity: while reviewing this I found #474.

phy1729:
> Also, 'f'oo (which was discussed on #468) is now actually highlighted as 'f' yellow+underline and oo default color + underline. Would it be possible for you to add a test for that as well? The zle internal combining logic might change in future zsh releases.

At present, no. The test harness only sees the `path` style for the whole of `'f'oo`. Some changes would need to be made to the test harness and API to specify multiple styles for a single `expected_region_highlight` element. Testing the merging happens as expected would be another issue on top.

danielshahaf:
Thanks. I think it's a behaviour we'd like to keep so I'll file a milestone=blue-sky issue about. Thanks for the merge. Excuse brevity.

danielshahaf:
> Thanks. I think it's a behaviour we'd like to keep so I'll file a milestone=blue-sky issue about. Thanks for the merge. Excuse brevity.

Filed #475.

PR #476, opened 2018-01-08
main: Highlight unclosed backtick subshells

phy1729:


highlighters/main/main-highlighter.zsh
```diff
@@ -986,6 +987,21 @@ _zsh_highlight_main_highlighter_highlight_dollar_quote()
   REPLY=$i
 }
 
+# Highlight backtick subshells
+_zsh_highlight_main_highlighter_highlight_backtick()
+{
+  local arg1=$1 i q=\` style
```
danielshahaf: arg1 is unused.

highlighters/main/main-highlighter.zsh
```diff
@@ -986,6 +987,21 @@ _zsh_highlight_main_highlighter_highlight_dollar_quote()
...
+  i=$arg[(ib:arg1+1:)$q]
```
danielshahaf: This doesn't account for escaped backticks, e.g., ```LBUFFER='echo `echo \`'``` and ```LBUFFER='echo `echo \``'``` (*sigh* markdown).  The double quote case does handle them; why does the difference?

highlighters/main/main-highlighter.zsh
```diff
@@ -986,6 +987,21 @@ _zsh_highlight_main_highlighter_highlight_dollar_quote()
...
+
+  if [[ $arg[i] == '`' ]]; then
+    style=back-quoted-argument
+  else
+    style=back-quoted-argument-unclosed
+  fi
+  _zsh_highlight_main_add_region_highlight $(( start_pos + $1 - 1 )) $(( start_pos + i )) $style
```
danielshahaf: You could use arg1 here.  It's more robust to use variable names without leading dollars; compare `s=1+1; echo $((2*s*2)) $((2*$s*2))`.

highlighters/main/main-highlighter.zsh
danielshahaf: @phy1729 points out it's used on the next line.

highlighters/main/main-highlighter.zsh
phy1729: Mostly because double quote already needed to handle `\` for `$` and history expansion. I can add the check.

danielshahaf:
``` echo `echo \`foo\`` ``` — in 0.6.0 the second word of the outer command is highlighted entirely as `back-quoted-argument`, with this patch the `foo` is unhighlighted even though it's part of the outer back-quoted-argument.

**Comment by phy1729** [comment]
Fixed in my testing and added to the test suite.

File: highlighters/main/main-highlighter.zsh [hunk 4]
```diff
@@ -986,6 +987,21 @@ _zsh_highlight_main_highlighter_highlight_dollar_quote()
   REPLY=$i
 }
 
+# Highlight backtick subshells
+_zsh_highlight_main_highlighter_highlight_backtick()
+{
+  local arg1=$1 i=$1 q=\` style
+  while i=$arg[(ib:i+1:)$q]; [[ $arg[i-1] == '\' && $(( i + 1 )) != $end_pos ]]; do done
```
danielshahaf: Nitpick: `!=`  → `-ne`

highlighters/main/test-data/back-quoted-argument.zsh
```diff
@@ -27,8 +27,9 @@
 # vim: ft=zsh sw=2 ts=2 et
 # -------------------------------------------------------------------------------------------------
 
-BUFFER='echo `echo 42`'
+BUFFER='echo `echo \`42\`` `echo 6 times 9'
```
danielshahaf: Suggest adding a command that the «42» is in command position, so if someone ever tries to support nested quotes and sees this test fail, it'll be clear to them that it's a false positive.

PR #482, opened 2018-01-13
main: Only highlight one base_style in _highlight_arguments

### LINKED ISSUE #481: default and path_prefix both set
```zsh
% tests/generate.zsh ': README' main x
% git diff
diff --git a/highlighters/main/test-data/x.zsh b/highlighters/main/test-data/x.zsh
index b93c1ec..8c72694 100644
--- a/highlighters/main/test-data/x.zsh
+++ b/highlighters/main/test-data/x.zsh
@@ -28,3 +28,10 @@
 # vim: ft=zsh sw=2 ts=2 et
 
+BUFFER=$': README'
+
+expected_region_highlight=(
+  '1 1 builtin' # :
+  '3 8 default' # README
+  '3 8 path_prefix' # README
+)
```

We shouldn't be adding _both_ default and path_prefix, should we?

phy1729:
Closes #481 

If this is ok, I think the check for `{single,double}-hyphen-option` should move into `_highlight_arguments` as a `base_style` as well.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -750,8 +750,8 @@ _zsh_highlight_main_highlighter_check_assign()
 _zsh_highlight_main_highlighter_highlight_path_separators()
 {
   local pos style_pathsep
-  style_pathsep=${style}_pathseparator
-  [[ -z "$ZSH_HIGHLIGHT_STYLES[$style_pathsep]" || "$ZSH_HIGHLIGHT_STYLES[$style]" == "$ZSH_HIGHLIGHT_STYLES[$style_pathsep]" ]] && return 0
+  style_pathsep=$1_pathseparator
+  [[ -z "$ZSH_HIGHLIGHT_STYLES[$style_pathsep]" || "$ZSH_HIGHLIGHT_STYLES[$1]" == "$ZSH_HIGHLIGHT_STYLES[$style_pathsep]" ]] && return 0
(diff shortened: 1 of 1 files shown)


phy1729:
Test failure seems to be a Travis issue.

danielshahaf:
Could you describe the change in words to help me review it? There seem to be two changes, around $reply and around $base_style?

phy1729:
Basic idea is helpers return highlights in $reply rather than calling _add_many_region_highlights themselves and then pick one of default, path, globbing as the base_style which is prior in region_highlight to the stuff the helpers return. Rebased to separate the changes into separate commits.

phy1729:
I think it's always been XPASSing on 5.2 and prior; haven't had time to look at why. (Although we could just merge #486 :smile:)

danielshahaf:
The travis tests for #486 pass which implies that that test point fails as expected (doesn't XPASS) there, on all zsh versions. So it's something in this PR that causes it to XPASS on 5.2 and earlier.

phy1729:
master has been failing since the test was added. https://travis-ci.org/zsh-users/zsh-syntax-highlighting/builds/329247186

phy1729:
#486 also makes the test that's currently XPASSing no longer a TODO.

danielshahaf:
I stand corrected.

danielshahaf:
> master has been failing since the test was added.

Okay. Branch LGTM. Please merge (before or after fixing that test in master, your call).

PR #486, opened 2018-01-18
Don't match redirection operators in command substitutions as <-> number range globs.

### LINKED ISSUE #483: $(<foo) is not globbing
`: $(<foo)` highlights the argument as a globbing. It isn't; it's a subcommand substitution that happens to contain a redirection.

Bisected to 987b743646c3e71a160de0ac2177bfe914bde39a.

danielshahaf:
Fixes #483.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -822,7 +822,7 @@ _zsh_highlight_main_highlighter_highlight_argument()
         elif [[ $arg[i+1] == [*@#?-$!] ]]; then
           (( i += 1 ))
         fi;;
-      [*?]*|\<*)
+      [*?]*|\<[0-9]#-[0-9]#\>)
 	# The '<' is for the <-> globbing syntax.  (This function doesn't get called on redirection tokens.)
         if $highlight_glob; then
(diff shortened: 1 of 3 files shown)


danielshahaf:
> This would break `foo<17-29>bar` (needs a trailing `*`).

Will fix.

> I'd prefer keeping the case on `$arg[i]` and only in the `<` case looking forward.

How would you implement lookahead in one case only?  Given that the `else` case would need to fall back/through to the default case it seemed cleaner to change the `case`'s needle/matchee.

phy1729:
```
+      \<)
+        if $highlight_glob && [[ ${arg:$i} =~ [0-9]*-[0-9]*\> ]]; then
+          (( i += $#MATCH ))
+          _zsh_highlight_main_add_region_highlight $start_pos $end_pos globbing
+          path_eligible=0
+          break
+        fi;;
```
(Of course making MATCH MBEGIN and MEND local).

Yes the change to `i` is currently useless, but #477 can use it.

danielshahaf:
https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:i483-v2 ?

danielshahaf:
Merged in 0458b0c16f7a.

PR #488, opened 2018-01-19
Post base_style enhancements

phy1729:


highlighters/main/main-highlighter.zsh
```diff
@@ -665,26 +665,16 @@ _zsh_highlight_highlighter_main_paint()
...
                  ;;
-        $'\x7d') # right brace
-                 #
-                 # Parsing rule: # {
-                 #
-                 #     Additionally, `tt(})' is recognized in any position if neither the
-                 #     tt(IGNORE_BRACES) option nor the tt(IGNORE_CLOSE_BRACES) option is set."""
-                 if $right_brace_is_recognised_everywhere; then
+        *)       if false; then
+                 elif [[ $arg = '}' ]] && $right_brace_is_recognised_everywhere; then
```
danielshahaf: The reason for spelling it `$'\x7d'` in the incumbent code is to avoid $EDITOR's brace-matching matching a syntactical `{` with a single-quoted `}`. That helps $EDITOR flag mismatched braces.

phy1729: Ah, I would suggest that this is a bug in $EDITOR and it could use better syntax-highlighting :wink:. Does `'}'` work? I don't have non-alphanumeric ASCII memorized.

danielshahaf: Then how about adding a named constant `readonly RIGHT_CURLY`? Or a comment on the line having the magic number?

danielshahaf: I'd rather work around such EDITOR behaviour than force maintainers to deal with it.

phy1729: Fair enough; comment it is.

highlighters/main/main-highlighter.zsh
```diff
@@ -905,6 +904,12 @@ _zsh_highlight_main_highlighter_highlight_double_quote()
     (( k = j + 1 ))
     case "$arg[$i]" in
       '"') break;;
+      '`') saved_reply=($reply)
+           _zsh_highlight_main_highlighter_highlight_backtick $i
+           (( i = REPLY ))
+           reply=($saved_reply $reply)
```
danielshahaf: It might be cleaner to have the function append to some variable other than `reply`, e.g., `prospective_reply`, throughout and then do once `reply=( "${prospective_reply[@]}" )` at the end, to avoid workarounds such as this one.

PR #491, opened 2018-01-20
'main': Highlight hyphen options as files when so-named files are present

danielshahaf:
E.g., «echo -foo» when «./-foo» exists.


phy1729:
I think this would be a regression.
```
touch -n
cat -n ~/.zshrc
```
Would highlight the `-n` as `path` not `single-hyphen-option`. `cat` doesn't care if a `-n` file exists; it interprets `-n` as an option.

It would be preferable for `cat -- -n` to highlight the `-n` as `path` since options may not occur after `--`.

danielshahaf:
Agree. I think my goal was to be clued to such files' existence, but that's not z-sy-h's job.

Issue #475, opened 2018-01-08
Test for path,single-quoted-string style combination


danielshahaf:
Currently,

```sh
% touch foo
% echo 'f'oo
```

highlights `'f'` in fg=yellow+underline and `oo` in underline. Add a unit test to make sure this behaviour is preserved.

See #473.

danielshahaf:
Test added.

The underlining no longer applies to the `'f'` part, though.

Issue #327, opened 2016-05-31
Make parser reusable


danielshahaf:
The `main` highlighter incorporates a zsh syntax parser.  Separate that parser into an independently reusable component.  The lexer is mostly handled by `bufferwords()` aka ${(z)}, although subshells and command substitutions are not currently recursed into (they're returned as a single token).

I do not have a concrete use-case for that at the moment, but I suspect one of these days somebody on zsh-workers@ will have an idea that will be much easier to implement were this done.

Note: it would be nice to reuse Src/parse.c's code, or to share code with compsys (`get_comp_string`).

danielshahaf:
Parsing parameter expansions should probably be its own project. :-)

PR #492, opened 2018-01-20
tests: Directly diff expected_region_highlight against region_highlight

phy1729:
Still has a few TODOs

- [x] comment-followed: Should the newline ending a comment be considered a `commandseparator`?
- [x] redirections2: What style should `>(wc)` be?
- [x] vanilla-newline: Unexpected entry `"0 0 unknown-token"`. Seems to be due to the last newline in `$PREBUFFER`.
- [x] vi-linewise-mode: I think that standout should be 1 15, but I'm not sure.
- [x] -unclosed styles are off by one.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -888,6 +888,8 @@ _zsh_highlight_main_highlighter_highlight_single_quote()
   if [[ $arg[i] == "'" ]]; then
     style=single-quoted-argument
   else
+    # If unclosed, i points past the end
+    (( i-- ))
     style=single-quoted-argument-unclosed
   fi
   reply=($(( start_pos + arg1 - 1 )) $(( start_pos + i )) $style $reply)
@@ -902,7 +904,7 @@ _zsh_highlight_main_highlighter_highlight_double_quote()
   local i j k style
   reply=()
 
-  for (( i = $1 + 1 ; i < end_pos - start_pos ; i += 1 )) ; do
+  for (( i = $1 + 1 ; i <= end_pos - start_pos ; i += 1 )) ; do
     (( j = i + start_pos - 1 ))
     (( k = j + 1 ))
     case "$arg[$i]" in
@@ -959,6 +961,8 @@ _zsh_highlight_main_highlighter_highlight_double_quote()
   if [[ $arg[i] == '"' ]]; then
     style=double-quoted-argument
   else
+    # If unclosed, i points past the end
+    (( i-- ))
     style=double-quoted-argument-unclosed
   fi
   reply=($(( start_pos + $1 - 1)) $(( start_pos + i )) $style $reply)
@@ -975,7 +979,7 @@ _zsh_highlight_main_highlighter_highlight_dollar_quote()
   integer c
   reply=()
 
-  for (( i = $1 + 2 ; i < end_pos - start_pos ; i += 1 )) ; do
+  for (( i = $1 + 2 ; i <= end_pos - start_pos ; i += 1 )) ; do
     (( j = i + start_pos - 1 ))
     (( k = j + 1 ))
     case "$arg[$i]" in
@@ -1011,6 +1015,8 @@ _zsh_highlight_main_highlighter_highlight_dollar_quote()
   if [[ $arg[i] == "'" ]]; then
     style=dollar-quoted-argument
   else
+    # If unclosed, i points past the end
+    (( i-- ))
     style=dollar-quoted-argument-unclosed
   fi
   reply=($(( start_pos + $1 - 1 )) $(( start_pos + i )) $style $reply)
@@ -1027,6 +1033,8 @@ _zsh_highlight_main_highlighter_highlight_backtick()
   if [[ $arg[i] == '`' ]]; then
     style=back-quoted-argument
   else
+    # If unclosed, i points past the end
+    (( i-- ))
     style=back-quoted-argument-unclosed
   fi
   reply=($(( start_pos + arg1 - 1 )) $(( start_pos + i )) $style)
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -104,7 +104,8 @@ _zsh_highlight_main_add_region_highlight() {
   (( start -= $#PREBUFFER ))
   (( end -= $#PREBUFFER ))
 
-  (( end < 0 )) && return # having end<0 would be a bug
+  (( start >= end )) && { print -r -- >&2 "zsh-syntax-highlighting: BUG: _zsh_highlight_main_add_region_highlight: start($start) >= end($end)"; return }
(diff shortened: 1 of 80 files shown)


danielshahaf:
Would this allow #475 to be implemented?

phy1729:
It couldn't test how ZLE combines styles, but it would ensure we know if we set `$region_highlight` differently.

danielshahaf:
> It couldn't test how ZLE combines styles, but it would ensure we know if 
> we set `$region_highlight` differently.

Good enough, I think.  It'd be nice to run an interactive zsh in a zpty and perform full-stack acceptance tests, but of course that's out of scope of this PR.

phy1729:
Tests pass now. There is still one `XXX` left which is point one above:
> comment-followed: Should the newline ending a comment be considered a `commandseparator`?

Ready for review.

danielshahaf:
Is there any `region_highlight`

phy1729:
Even with `ZSH_HIGHLIGHT_STYLES[commandseparator]=bg=cyan,fg=cyan,bold,standout,underline`, which seems to be every type of highlighting ZLE supports, I don't see anything for
```
: foo
bar
```
Although that newline is highlighted as `commandseparator`. I suppose this means the issue is purely academic until #327.

danielshahaf:
To answer your question, I think the newline following a comment _should_ be a commandseparator, to maintain the invariant that the token preceding a command position (a physical one, i.e., not a synthetic one after a precommand or assignment or the keyword `if`) is always `commandseparator`.

I also agree that it's an academic concern until #327 (and as such not a blocker).

phy1729:
Filed #501 and changed the test to XFAIL. Rebased onto master.

highlighters/main/main-highlighter.zsh
```diff
@@ -104,7 +104,8 @@ _zsh_highlight_main_add_region_highlight() {
   (( start -= $#PREBUFFER ))
   (( end -= $#PREBUFFER ))
 
-  (( end < 0 )) && return # having end<0 would be a bug
+  (( start >= end )) && { print -r -- >&2 'zsh-syntax-highlighting: BUG: _zsh_highlight_main_add_region_highlight: start >= end'; return }
```
danielshahaf: Suggest to interpolate `"$start"` and `"$end"` into the error message.

(Generally, most error message should have something interpolated into them.)

highlighters/main/main-highlighter.zsh
```diff
@@ -104,7 +104,8 @@ _zsh_highlight_main_add_region_highlight() {
...
+  (( start < 0 && end <= 0 )) && return
```
danielshahaf: At this point, `end <= 0` implies `start < 0` so the first conjunct is redundant.

highlighters/main/test-data/alias.zsh
```diff
@@ -43,7 +43,7 @@ BUFFER='x.alias2; alias1'
 # functionality is present, and skip verifying suffix-alias highlighting
 # if it isn't.
 expected_region_highlight=()
-if [[ "$(type -w x.alias2)" == *suffix* ]]; then
+if zmodload -e zsh/parameter || [[ "$(type -w x.alias2)" == *suffix* ]]; then
```
danielshahaf: Nice!

tests/test-highlighting.zsh
```diff
@@ -106,62 +106,56 @@ run_test_internal() {
...
+    unset expected_highlight_zone
+    unset exp_start exp_end
     unset todo
     unset highlight_zone
+    unset start end
+    unset desc
   done
+
+  if (( $#expected_region_highlight == $#region_highlight )); then
+    print -r -- "ok $i - cardnality check"
```
danielshahaf: "cardinality" is misspelled.

highlighters/main/test-data/double-quoted2.zsh
```diff
@@ -32,6 +32,8 @@
 BUFFER=': "foo$bar'
 
 expected_region_highlight=(
-  "3 6 double-quoted-argument-unclosed" # "foo
+  "1 1 builtin" # :
+  "3 10 default" # "foo$bar
+  "3 10 double-quoted-argument-unclosed" # "foo$bar
```
danielshahaf: For reference, this double highlighting is the subject of #495.

tests/test-highlighting.zsh
```diff
@@ -106,62 +106,56 @@ run_test_internal() {
...
-      done
-    else
-      # noop range; ignore.
-    fi
-    unset start end
-    unset highlight_zone
-  done
+  if (( unsorted )); then
+    region_highlight=("${(@n)region_highlight}")
+    expected_region_highlight=("${(@n)expected_region_highlight}")
```
danielshahaf: Suppose the input contains both `42 foo` and `42 bar`, is numeric sorting guaranteed to always output them in the same order, regardless of what other elements the input contains?

Plain old lexicographic sorting would definitely have this property but I suspect it would make for less nice failure modes (I presume we'd want failure to be reported on the "first" non-match, for some notion of "match" that depends on the start,stop indices).

highlighters/main/main-highlighter.zsh
```diff
@@ -888,6 +888,8 @@ _zsh_highlight_main_highlighter_highlight_single_quote()
   if [[ $arg[i] == "'" ]]; then
     style=single-quoted-argument
   else
+    # If unclosed, i points past the end
+    (( i-- ))
```
danielshahaf: If I revert this line, `multiline-string2` fails.  👍

danielshahaf:
Thank you for making this happen!

I've went through the diff and added a few comments.  I haven't the brainwidth to review all the new logic around byte/character offsets, though; I'm going to go ahead and trust you on it.

tests/test-highlighting.zsh
phy1729:
```
% a=(1 3 '2 foo' '2bar' '2 2 bar' '2 14 bar' '21 foo'); printf '<%s>' ${(@n)a} 
<1><2 2 bar><2 14 bar><2 foo><2bar><3><21 foo>
```

So it looks like numeric sorting falls back to lexicographic sorting.

danielshahaf: The implementation of the `(n)` sorting flag (NUMERIC_GLOB_SORT option) is:

https://github.com/zsh-users/zsh/blob/567733906210feaebce94675985eb1e77602bfd8/Src/sort.c#L119-L143

phy1729: In English:
1) Skip characters that are equal.
2) Store the comparison between the two different chars (This is how (n) falls back to lexicographical).
3) Assuming one of the differing characters is a digit, rewind to the start of digits.
4) Skip `0`s.
5) Skip equal digits.
6) Store the difference.
7) Find length and return in favor of the longer or the stored difference if equal.

So I think `(n)` DTRT.

PR #493, opened 2018-01-20
'main': Don't highlight bare '$foo' as a filename, as it's a parameter expansion.

### LINKED ISSUE #474: path wrongly precedes dollar
Similar to #468, but unlike it, not a regression:

```zsh
% touch \$foo
% echo $foo \$foo
```

shouldn't highlight the first `$foo` as a path.

danielshahaf:
Fixes #474.


phy1729:
:+1:

Issue #362, opened 2016-09-07
Decide on a zstyle structure


danielshahaf:

phy1729:
It may also be desirable to include the nesting level; so that if commands in `$( )` are highlighted, they may be highlighted differently.

phy1729:
It would be nice to have a consistent way to distinguish between a completed element and an in progress one (e.g. path and path_prefix; c.f. #148 #373).

danielshahaf:
> It would be nice to have a consistent way to distinguish between a completed element and an in progress one (e.g. path and path_prefix; c.f. #148 #373).

tldr:

#148 - special highlighting for the word currently being typed, e.g., for the command word depending on its possible completions

#373 - highlighting of unterminated string literals

phy1729:
It may be useful to make the style reflect the nesting mentioned in #495 e.g. `--"delete-${when}"` the `--` would be `...:option-hyphen-double`, `"delete-` `...:option-hyphen-double:quoted-double`, and `${when}` `...:option-hyphen-double:quoted-double:parameter`. Not sure if we should no longer overlap styles if we do that.
(Of course the particular names aren't important, just giving an example.)

PR #496, opened 2018-01-21
main: Add ZSH_HIGHLIGHT_DIRS_BLACKLIST

### LINKED ISSUE #379: Disable path checking by directory
RFE: Disable path checking for some directories but not others.  Similar to the `vcs_info` [`disable-patterns` style](http://zsh.sourceforge.net/Doc/Release/User-Contributions.html#index-disable_002dpatterns).  This would be useful under expensive-to-readdir()/stat() mountpoints (e.g., sshfs).

This would need to check several things:
- [ ] Absolute filenames under DIR: `% stat /victim/foo/bar`
- [ ] Relative paths that resolve to paths under DIR (could use the `:a` modifier here, since it's disk-agnostic): `% stat ../victim/foo/bar`
- [ ] Any paths, when cwd is DIR:  `/victim% stat foo` (where the prompt displays cwd)
- [ ] Any of the foregoing, as a command word: `% /victim/foo/bar`

phy1729:
Closes #379.

This may fail on symlinks, but I think this should cover most cases. I placed the check below the -L and -e tests so once the file is completely typed it will still highlight as path.

_Please test_ that the speed with this change is usable if you've run into #379. If things are still too slow the check can be moved up (or we could test that the word is no longer being typed with `$CURSOR` and have a cache).

I'll add docs after we get some reports and know how `$ZSH_HIGHLIGHT_DIRS_BLACKLIST` will act.

### DIFF
--- a/docs/highlighters/main.md
+++ b/docs/highlighters/main.md
@@ -71,6 +71,12 @@ The syntax for values is the same as the syntax of "types of highlighting" of
 the zsh builtin `$zle_highlight` array, which is documented in [the `zshzle(1)`
 manual page][zshzle-Character-Highlighting].
 
+#### Parameters
+
+To avoid partial path lookups on a path, add the path to the `ZSH_HIGHLIGHT_DIRS_BLACKLIST` array.
+
+    ZSH_HIGHLIGHT_DIRS_BLACKLIST+=(/mnt/slow_share)
+
 ### Useless trivia
 
 #### Forward compatibility.
--- a/docs/highlighters/main.md
+++ b/docs/highlighters/main.md
@@ -73,9 +73,10 @@ manual page][zshzle-Character-Highlighting].
 
 #### Parameters
 
-To avoid partial path lookups on a path, add the path to the `ZSH_HIGHLIGHT_DIRS_BLACKLIST` array.
+To avoid partial path lookups on a path, add the path to the `X_ZSH_HIGHLIGHT_DIRS_BLACKLIST` array.
+This interface is still experimental.
 
-    ZSH_HIGHLIGHT_DIRS_BLACKLIST+=(/mnt/slow_share)
+    X_ZSH_HIGHLIGHT_DIRS_BLACKLIST+=(/mnt/slow_share)
 
 ### Useless trivia
 
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -757,13 +757,26 @@ _zsh_highlight_main_highlighter_highlight_path_separators()
 _zsh_highlight_main_highlighter_check_path()
 {
   _zsh_highlight_main_highlighter_expand_path $arg;
-  local expanded_path="$REPLY"
+  local expanded_path="$REPLY" tmp_path
 
   REPLY=path
 
   [[ -z $expanded_path ]] && return 1
   [[ -L $expanded_path ]] && return 0
   [[ -e $expanded_path ]] && return 0
+
+  # Check if this is a blacklisted path
+  if [[ $expanded_path[1] == / ]]; then
+    tmp_path=$expanded_path
+  else
+    tmp_path=$PWD/$expanded_path
+  fi
+  tmp_path=$tmp_path:a
+
+  while [[ $tmp_path != / ]]; do
+    [[ -n "${(M)ZSH_HIGHLIGHT_DIRS_BLACKLIST:#$tmp_path}" ]] && return 1
+    tmp_path=$tmp_path:h
+  done
 
   # Search the path in CDPATH
   local cdpath_dir
@@ -1070,3 +1083,4 @@ else
   # Make sure the cache is unset
   unset _zsh_highlight_main__command_type_cache
 fi
+typeset -ga ZSH_HIGHLIGHT_DIRS_BLACKLIST
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -774,7 +774,7 @@ _zsh_highlight_main_highlighter_check_path()
   tmp_path=$tmp_path:a
 
   while [[ $tmp_path != / ]]; do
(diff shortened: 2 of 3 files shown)


phy1729:
Seems `${name:|arrayname}` was added in zsh 5.0.0.

phy1729:
Merging this to master as an experimental interface. Hopefully this will make testing easier. If positive feedback ensues, the `X_` can be dropped.

phy1729:
We got some feedback in #528 and moved the blacklist check to prior to any syscalls. @danielshahaf do you think that is sufficient to drop the `X_` prefix before 0.7.0 or should we leave it this round?

danielshahaf:
@phy1729 Let's see. The feature was added in 0.6.0-89-g6713727 (67137277420061d396900d8a7c85f1a444112820).  On the one hand, that's 18 months ago (a very long time, so far as exposure to users of `master` is concerned); on the other hand, the feature hasn't appeared in any tagged release yet.  Both the functionality and the implementation are straightforward and unlikely to change.

In light of all these, I guess we can just drop the `X_` prefix?  We can always do a 0.7.0-rc1 release to solicit some more feedback.  (Actually, we can tag rc1 right now, can't we?)

If you want to make it a zstyle, by the way, that'd work for me — see #362; so long as we put a version number or a format number at the head of the style, that'd raise no forward compatibility issues — but of course, simply renaming the parameter is a lot easier.

What will the transition be? Will the code continue to honour `$X_ZSH_HIGHLIGHT_DIRS_BLACKLIST` if it is set?  What about if `$X_ZSH_HIGHLIGHT_DIRS_BLACKLIST` and `$ZSH_HIGHLIGHT_DIRS_BLACKLIST` are both set?

phy1729:
I would leave changing to a zstyle to later when the rest of the config will be changed as well.

Regarding the transition, I was thinking something like the patch below. The notice in zsh-syntax-highlighting.zsh will be nicer to hit as it's prior to drawing the prompt. The notice in main-highlighter.zsh will catch if `X_ZSH_HIGHLIGHT_DIRS_BLACKLIST` is set after we'd sourced. In either case the code configured the new variable, so the blacklist continues working.

I would like to put this (and the X_ removal) in now, tag -rc1, and then remove the warnings prior to release in a week. Those who follow master tend to upgrade frequently. Although, I have no objection to lengthening the amount of time this stays in to catch anyone on vacation or otherwise AFK.

```diff
diff --git a/highlighters/main/main-highlighter.zsh b/highlighters/main/main-highlighter.zsh
index 91be6ca..838ab2c 100644
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -355,6 +355,12 @@ _zsh_highlight_highlighter_main_paint()
     '!' # reserved word; unrelated to $histchars[1]
   )
 
+  if (( $+X_ZSH_HIGHLIGHT_DIRS_BLACKLIST )); then
+    print >&2 'X_ZSH_HIGHLIGHT_DIRS_BLACKLIST is deprecated. Please use ZSH_HIGHLIGHT_DIRS_BLACKLIST.'
+    ZSH_HIGHLIGHT_DIRS_BLACKLIST=($X_ZSH_HIGHLIGHT_DIRS_BLACKLIST)
+    unset X_ZSH_HIGHLIGHT_DIRS_BLACKLIST
+  fi
+
   _zsh_highlight_main_highlighter_highlight_list -$#PREBUFFER '' 1 "$PREBUFFER$BUFFER"
 
   # end is a reserved word
diff --git a/zsh-syntax-highlighting.zsh b/zsh-syntax-highlighting.zsh
index df8a1aa..6b75a5e 100644
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -432,6 +432,12 @@ zmodload zsh/parameter 2>/dev/null || true
 # Initialize the array of active highlighters if needed.
 [[ $#ZSH_HIGHLIGHT_HIGHLIGHTERS -eq 0 ]] && ZSH_HIGHLIGHT_HIGHLIGHTERS=(main)
 
+if (( $+X_ZSH_HIGHLIGHT_DIRS_BLACKLIST )); then
+  print >&2 'X_ZSH_HIGHLIGHT_DIRS_BLACKLIST is deprecated. Please use ZSH_HIGHLIGHT_DIRS_BLACKLIST.'
+  ZSH_HIGHLIGHT_DIRS_BLACKLIST=($X_ZSH_HIGHLIGHT_DIRS_BLACKLIST)
+  unset X_ZSH_HIGHLIGHT_DIRS_BLACKLIST
+fi
+
 # Restore the aliases we unned
 eval "$zsh_highlight__aliases"
 builtin unset zsh_highlight__aliases
```

danielshahaf:
> I would leave changing to a zstyle to later when the rest of the config will be changed as well.

+1

> I would like to put this (and the X_ removal) in now, tag -rc1, and then remove the warnings prior to release in a week. Those who follow master tend to upgrade frequently. Although, I have no objection to lengthening the amount of time this stays in to catch anyone on vacation or otherwise AFK.

That all sounds good, with the caveat you foresaw: I don't know if leaving it in for just a week would be enough. How often do those plugin managers auto-update us?



> +  if (( $+X_ZSH_HIGHLIGHT_DIRS_BLACKLIST )); then

Should we check not only whether the array is defined but also whether it's not empty?  I guess the condition is good as-is, but thought I'd ask.

> +    print >&2 'X_ZSH_HIGHLIGHT_DIRS_BLACKLIST is deprecated. Please use ZSH_HIGHLIGHT_DIRS_BLACKLIST.'

Let's prefix the output with `zsh-syntax-highlighting: `.

> +    ZSH_HIGHLIGHT_DIRS_BLACKLIST=($X_ZSH_HIGHLIGHT_DIRS_BLACKLIST)

What if both variables are set?  I don't think we should just overwrite the value of `$ZSH_HIGHLIGHT_DIRS_BLACKLIST` in that case.

> +    unset X_ZSH_HIGHLIGHT_DIRS_BLACKLIST

Thanks.

PR #502, opened 2018-02-11
main: Do not highlight options after --

phy1729:
Re #491, something like this? It does incorrectly highlight `-n` as path in a case statement like
```
case a in --) foo;; a) cat -n .zshrc;; esac
```
(Should be `single-hyphen-option`; is highlighted as `path`.)


danielshahaf:
@phy1729 This is milestoned 0.8.0.  Do you intend to work on it for the 0.8.0 release?  If yes, great; if not, could you re-milestone it appropriately?  Cheers.

danielshahaf:
(Assigning until the question is answered. Feel free to unassign yourself if needed.)

danielshahaf:
Triage: Bumping to the next milestone due to feedback timeout.

PR #503, opened 2018-02-11
main: Remove quotes before tilde expansion

phy1729:
`ls \~` does not expand; however, `ls ~/"user"` does. Test that the first character is `~` before dequoting.

I couldn't figure out why the assignment is done in a PE. If there's a reason, please enlighten me so a comment can be added.


danielshahaf:
The comment on line 1050 hints at the answer: we don't want to perform globbing (because that can execute code)

phy1729:
Seems to be for any assignment
```
% print -r ${~${:-'~/?'}}                          
/home/phy1729/* /home/phy1729/< /home/phy1729/@ /home/phy1729/Q
% a=${~${:-'~/?'}}; typeset -p a
typeset a='/home/phy1729/?
```

danielshahaf:
I'm sure I tested this at the time, but not what I contrasted `${:=}` with. I can't check `blame` here.

phy1729:
Your change was 5cd1fb2 and the assignment PE was already present. 8d4cde6 was the commit that introduced the assignment PE.

danielshahaf:
Thanks. 8d4cde6 contrasts freestanding and RHS of `:=` so that might be the comment's meaning.

PR #508, opened 2018-03-10
path_prefix is not used when PREBUFFER is not empty

phy1729:
`end_pos` is relative to `buf` which is `$PREBUFFER$BUFFER`. Thus the check should be if we're at the end of `buf` not `BUFFER`.

### DIFF
--- a/highlighters/main/test-data/path_prefix3.zsh
+++ b/highlighters/main/test-data/path_prefix3.zsh
@@ -1,0 +1,38 @@
+# -------------------------------------------------------------------------------------------------
+# Copyright (c) 2015 zsh-syntax-highlighting contributors
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are permitted
+# provided that the following conditions are met:
+#
+#  * Redistributions of source code must retain the above copyright notice, this list of conditions
+#    and the following disclaimer.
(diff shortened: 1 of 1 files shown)


danielshahaf:
+1.

The check for `"-"` was added in f26310d97e41d165b6ba3ce85940ccfeb5c8b35b and there are multiple issues with it: it ignores $PREBUFFER, it ignores the possibility of command separators (`ZSH_HIGHLIGHT_TOKENS_COMMANDSEPARATOR`), and I don't understand the need for it in the first place...  why wouldn't `path_prefix` be shown when `-` is prefixed to argv[0] of a launched command?

phy1729:
Perhaps it was to avoid highlighting options as path or path_prefix? That is now handled by `_zsh_highlight_main_highlighter_highlight_argument`. So perhaps we can drop `${BUFFER[1]} != "-" &&` now.

danielshahaf:
> Perhaps it was to avoid highlighting options as path or path_prefix? 

${BUFFER[1]} is the first letter of $BUFFER, not the first letter of the current word.

> That is now handled by 
> `_zsh_highlight_main_highlighter_highlight_argument`. So perhaps we can 
> drop `${BUFFER[1]} != "-" &&` now.

Yeah, perhaps.

I'd ping Guido to ask if he remembers the use case, but the code is seven years old...

Issue #494, opened 2018-01-21
Highlight process substitutions


phy1729:
Highlight `<(list)`, `>(list)`, and `=(list)` as `process-substitution`. Related to #139.

Issue #202, opened 2015-09-22
command word: non-executable file should be unknown-token


danielshahaf:
The following:

``` zsh
touch foo
chmod 0644 foo
./foo
```

highlights `./foo` as `path`, but should highlight it as `unknown-token` (unless there happens to be an alias or function literally called `./foo`).

Note that we already call `stat` once (since `./does-not-exist` is highlighted as `unknown-token`).

danielshahaf:
`_zsh_highlight_main_highlighter_check_path` needs to know whether the current word is at command word position, and only require check executability when at command word.

danielshahaf:
Directories may be the command word when the `AUTO_CD` option is set.  (However, a patch that fixes the issue only for the common case of `NO_AUTO_CD` would be accepted.)

danielshahaf:
In `unset x; $x ls`, the `ls` is also a command word.

danielshahaf:
> In `unset x; $x ls`, the `ls` is also a command word.

PR #667.

danielshahaf:
... and PR #669 for the other issues here.

PR #512, opened 2018-03-28
Highlight command and process substitution

### LINKED ISSUE #139: backticks and $(...) aren't highlighted properly
Say, in this command:
`vim $(find -name "something")`
both `vim` and `find` should be highlighted as a command, but only `vim` is.

phy1729:
Closes #139 and #494.

Needs naming suggestions for things with `XXX`, but should be otherwise ready,


phy1729:
Tests only fail for 5.0.8.

`command-substitution-unclosed` fails with 
```
not ok 2 - [3,16] «foo$(echo bar» - expected (3 15 "default"), observed (3 16 "default").
```
and `process-substitution2` with
```
not ok 6 - [17,27] «=(echo foo» - expected (17 26 "default"), observed (17 27 "default").
```

Not sure why yet.

danielshahaf:
> Seems `${(z)` in 5.0.8 splits `: foo$(echo bar` into `:` and `foo$(echo bar ` (note the extra space at the end). Since this is a zsh bug that's already been fixed and will nearly never* effect the highlighting the user sees (since the off by one will almost always be past the end of the buffer), can we just add a condition in the tests to expect the wrong output on zsh 5.0.8?
> 

Sure.  Making the condition dependent not on $ZSH_VERSION but on [[ ${${(z):-': foo$(echo bar'}[2]} == *' ' ]] (sp?) would be nice, if it's not much effort.

> *: Nearly never because it's possible that a user could nest an unclosed `$(` command substitution inside of a backtick command substitution.

So... If someone nests a $() inside backticks under 5.0.8, their highlighting will be off-by-one until they type the closing paren?  I think we can live with that :-)

Cheers,

phy1729:
> Making the condition dependent not on $ZSH_VERSION but on [[ ${${(z):-': foo$(echo bar'}[2]} == *' ' ]] (sp?) would be nice, if it's not much effort.

It wouldn't be hard, but I think it would be better to check $ZSH_VERSION to catch possible regressions in the future (hopefully before they become part of a release). The latest commit reflects this although the condition would be easy to change if you still think the other test is better.

danielshahaf:
To catch regressions in ${(z)} is zsh's test suite's business, not ours. I vote for the functional test here and a regression test in zsh's test suite. Makes sense?

phy1729:
Works for me. Rebased.

danielshahaf:
Ty. Still planning to review/test once I'm at my workstation. Haven't had time to breathe.

danielshahaf:
Reviewed. Haven't tested (not at workstation). Kudos for the work & thanks for patience.

Does zsh have a regression test for that 5.0.8 bug?

danielshahaf:
XXX -> parse_list? Using the zsh grammar meaning of "list".

phy1729:
Not quite correct since the function can highlight multiple lists. I don't think zsh has a name for one or more lists.

danielshahaf:
What do you mean by "multiple lists"?

A list can contain other lists, e.g., a list that contains an `if` construct.

> More generally, a list can be seen as a set of any shell commands whatsoever, including the complex commands below; this is implied wherever the word `list' appears in later descriptions.  For example, the commands in a shell function form a special sort of list.

highlighters/main/test-data/back-quoted-argument.zsh
```diff
@@ -27,16 +27,25 @@
...
+  "7 10 builtin" # echo
+  "12 17 default" # \`42\`
+  "12 17 back-quoted-argument" # \`42\`
+  "14 15 unknown-token" # 42
   "20 39 default" # "is `echo equal` to"
   "20 39 double-quoted-argument" # "is `echo equal` to"
   "24 35 back-quoted-argument" # `echo equal`
+  "25 28 builtin" # echo
+  "30 34 default" # equal
   "41 55 default" # `echo 6 times 9
```
danielshahaf: Should the backticks themselves be highlighted?  Quotes and command/process substitutions are all highlighted.  I suppose we could highlight backticks as magenta, like `$(…)`, since both of these are subject to word splitting.

highlighters/main/main-highlighter.zsh
```diff
@@ -1083,21 +1083,67 @@ _zsh_highlight_main_highlighter_highlight_dollar_quote()
...
-  fi
-  reply=($(( start_pos + arg1 - 1 )) $(( start_pos + i )) $style)
+  last=$(( arg1 + 1 ))
+  # Remove one layer of backslashes and find the end
+  while i=$arg[(ib:i+1:)[\\\\\`]]; do # find the next \ or `
+    if (( i > end_pos - start_pos )); then
+      buf=$buf$arg[last,i]
+      offsets[i-arg1-offset]='' # So we never index past the end
+      (( i-- ))
+      subshell_has_end=$(( has_end && (start_pos + i == len) ))
```
danielshahaf: No tests fail if I change `has_end &&` to `0 &&`.  How about adding this one:

```diff
diff --git a/highlighters/main/test-data/xxx.zsh b/highlighters/main/test-data/xxx.zsh
new file mode 100644
index 0000000..d73fc14
--- /dev/null
+++ b/highlighters/main/test-data/xxx.zsh
@@ -0,0 +1,40 @@
+#!/usr/bin/env zsh
+# -------------------------------------------------------------------------------------------------
+# Copyright (c) 2018 zsh-syntax-highlighting contributors
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are permitted
+# provided that the following conditions are met:
+#
+#  * Redistributions of source code must retain the above copyright notice, this list of conditions
+#    and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright notice, this list of
+#    conditions and the following disclaimer in the documentation and/or other materials provided
+#    with the distribution.
+#  * Neither the name of the zsh-syntax-highlighting contributors nor the names of its contributors
+#    may be used to endorse or promote products derived from this software without specific prior
+#    written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
+# IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+# FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+# IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
+# OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# -------------------------------------------------------------------------------------------------
+# -*- mode: zsh; sh-indentation: 2; indent-tabs-mode: nil; sh-basic-offset: 2; -*-
+# vim: ft=zsh sw=2 ts=2 et
+# -------------------------------------------------------------------------------------------------
+
+touch foo
+BUFFER=$': `ls fo'
+
+expected_region_highlight=(
+  '1 1 builtin' # :
+  '3 8 default' # `ls fo
+  '3 8 back-quoted-argument-unclosed' # `ls fo
+  '4 5 command' # ls
+  '7 8 path_prefix' # fo
+)
```

danielshahaf:
Reviewed.  Great work!  Thanks for splitting the noop changes to their own commits.

danielshahaf:
After `touch foo`, `BUFFER='echo $(foo)' paints `foo` in magenta+underline, whereas I'd expect it to be highlighted as unknown-token since it's in command position.

phy1729:
> After `touch foo`, `BUFFER='echo $(foo)'` paints `foo` in magenta+underline, whereas I'd expect it to be highlighted as unknown-token since it's in command position.

Similar for `touch foo` then `BUFFER=foo`, so I think it's a separate issue.

danielshahaf:
I don't see the similarity: here, `BUFFER=foo` has nothing in magenta.  (The filename _is_ underlined, though :smile:)

To be clear, in my example, I'd typed `echo $(foo)` at the prompt; I did not literally type `BUFFER=` at the prompt.

highlighters/main/test-data/back-quoted-argument.zsh
phy1729: That wouldn't be consistent with single, double, and dollar quotes, but then again those are quotes and these are command substitutions.

danielshahaf:
Ah, fair enough.

At the risk of shedding bikes, should the whole of `$(foo)` be highlighted in magenta, or just the delimiters (`$(` and `)`)?

highlighters/main/test-data/back-quoted-argument.zsh
danielshahaf: _What_ wouldn't be consistent? The colour (yellow v. magenta), or something else?

phy1729: For the quoting styles, the quotes themselves are not given a special style. If `$(` and `)` and backticks are, that would be inconsistent between the two; but they are different syntactical constructs. Additionally I think magenta (or any other really) highlighting all the words not otherwise highlighted in a command substitution is a bit much.

danielshahaf:
> Oh, the magenta is because the whole $(foo) is highlighted as command-substitution. I was intending to convey once BUFFER=foo is highlighted as unknown-token so will echo $(foo).

That's #202.

highlighters/main/test-data/back-quoted-argument.zsh
danielshahaf: Agreed that highlighting all words inside a cmdsubst as magenta is a bit much.  How about highlighting `$(`, `)`, `` ` `` as `command-substitution-delimiter`, defaulting to magenta, and keeping plain words inside the cmdsubst unhighlighted?  We could even keep the `command-substitution` group matching the entire `$(…)` construct so people could assign to it in themes.

An orthogonal question: if we had `command-substitution-{delimiter-,}$(( N++ ))` styles, with $N being the nesting depth, we could have a [rainbow parentheses](https://github.com/luochen1990/rainbow) theme.  Featuritis?

phy1729: I was thinking we'd consider that when coming up with zstyles for #362 . I like the delimiter idea. How about the latest two commits?

Note: `: "$(: foo)"` will still have foo highlighted yellow since there's the double-quoted-argument highlighting applying to the whole of the double quotes. Not sure what if anything to do about that.

danielshahaf: Delimiters LGTM.

Re `foo` in `: "$(: foo)"`, I don't think it should be yellow since the lexing rules of double-quoted strings (e.g., backslash escape sequences) don't apply there. That's not to say whether or not it blocks merging or releasing; just to say that I don't think it's useful for that `foo` to be yellow.

highlighters/main/main-highlighter.zsh
danielshahaf: Ping?

phy1729: LGTM. I should probably make a second pass and see if any more test ideas come to mind.

danielshahaf: Thanks for the review.  Want to append it to the branch?

highlighters/main/test-data/back-quoted-argument.zsh
phy1729: I agree, but I don't know what to do about it. In theory
`: ${ZSH_HIGHLIGHT_STYLES[default]:=fg=default}`
would help, but it appears to make no difference here. Additionally the yellow would still show through for `path` styled strings.

highlighters/main/main-highlighter.zsh
```diff
@@ -584,12 +618,22 @@ _zsh_highlight_highlighter_main_paint()
...
+                          _zsh_highlight_main_add_region_highlight $start_pos $end_pos $style
+                          already_added=1
+                          local i=$(( arg[(i)=] + 1 ))
+                          if [[ $arg[i] == '(' ]]; then
                             in_array_assignment=true
                           else
                             # assignment to a scalar parameter.
                             # (For array assignments, the command doesn't start until the ")" token.)
                             next_word+=':start:'
+                            if (( i <= end_pos )); then
```
phy1729: This condition is incorrect. Should be 
```(( start_pos + i <= end_pos ))```
Fixed up in latest branch.

**Comment by danielshahaf** [comment]
> I agree, but I don't know what to do about it. In theory
> `: ${ZSH_HIGHLIGHT_STYLES[default]:=fg=default}`
> would help, but it appears to make no difference here. Additionally the 
> yellow would still show through for `path` styled strings.

Two things come to mind:

1. zle has some issues with composing multiple styles for the same offsets, e.g., I think `region_highlight=("1 2 fg=foo" "1 2 bg=bar")` doesn't results in fg=foo,bg=bar but in bg=bar.  (I'm not sure about this specific example but that's the gist.)  Depending on how you tested you may have run into that.  That'll have to be fixed in zle.

2. We could break down the double-quoted string and only set fg=yellow for the parts outside of any inner command substitutions, e.g., in «: "foo `bar` baz"» we'd add separate ranges for the «foo » and « baz».  I think that's what pws would describe as a finite amount of effort ;-).

**Comment by danielshahaf** [comment]
> phy1729 commented on this pull request.
> This condition is incorrect. Should be 
> ```(( start_pos + i <= end_pos ))```
> Fixed up in latest branch.

Regression test?

**Comment by phy1729** [comment]
As the branch is `A[a=1]=2` is highlighted as
```
0 8 assign
4 8 default
```
This is incorrect since the second style should start at 7.

Although `A[a=$(echo 1)]=2` is valid and results in `a=1; A=(2)`. Not sure what the correct behavior is. Perhaps
```
0 16 assign
2 13 assign
4 13 default
4 13 command-substitution
6 10 command
11 12 default
15 16 default
```

danielshahaf:
That `$(echo)` should be handled like either `: $(( $(echo) ))`, I presume?  Subscripts are similar to arithmetic expressions, parsing-wise.

To make things even more fun, extra `=` signs may occur in the RHS as well: `A[a=1]=$B[b=2]`

phy1729:
I don't know if we can correctly find the end of the subscript without adding an arithmetic expression parser which I think is out of scope for this PR.

Issue #506, opened 2018-02-26
There is a problem when i login my root account


Alienover:
> zsh-syntax-highlighting: 'README.md' highlighter should define both required functions '_zsh_highlight_highlighter_README.md_paint' and '_zsh_highlight_highlighter_README.md_predicate' in '/Users/jiarong/.oh-my-zsh/custom/plugins/zsh-syntax-highlighting/highlighters/README.md/README.md-highlighter.zsh'.

After the configuration, when i login my root account in the iterm, this note show up. 

Does anybody come up with this problem?

phy1729:
It sounds like you have something similar to the following in your config
```
ZSH_HIGHLIGHT_HIGHLIGHTERS=(zsh-syntax-highlighting/highlighters/*(:t))
```
which will enable all files in the highlighters directory as a highlighter. That directory also contains a `README.md` file which is causing the issue. Adding only directories
```
ZSH_HIGHLIGHT_HIGHLIGHTERS=(zsh-syntax-highlighting/highlighters/*(/:t))
```
would work around the issue; however, it would be better to explicitly list the highlighters you want in the order you want.

Also you should consider not using a root shell and using `sudo` instead.

phy1729:
Closing due to inactivity.

NearAutomata:
@phy1729 Your suggested solution is a patchwerk approach that doesn't fix the underlying problem in the first place. I suggest this should be actually fixed in the plugin.

danielshahaf:
The erroneous code is not in our git, it is in users' zshrc files.  What change to z-sy-h's code would you suggest?

@phy1729 I would add the `-` glob qualifier before the `/` to allow symlinks to directories, too.

NearAutomata:
@danielshahaf I am using oh-my-zsh and including the plugin in my zshrc via `plugins=(zsh-syntax-highlighting)` and it started to write the warning just recentlyish. Personally I don't know how to fix it in the first place.

danielshahaf:
oh-my-zsh is a separate project.  File a bug with them (there isn't an existing one).  Excuse brevity.

danielshahaf:
To be clear: there is no code in tip of master of this repository that sets ZSH_HIGHLIGHT_HIGHLIGHTERS to anything.  I don't know where the erroneous code @phy1729 describes is, but it's not in this repository.

PR #518, opened 2018-05-30
fix _zsh_highlight_load_highlighters when executing zsh with sudo rights

qoomon:
I am facing following error message if i start my zsh with sudo right e.g. `sudo zsh`

**Error Log**
_zsh-syntax-highlighting: 'README.md' highlighter should define both required functions '_zsh_highlight_highlighter_README.md_paint' and '_zsh_highlight_highlighter_README.md_predicate' in '/Users/USERNAME/.zsh.zgem/gems/zsh-syntax-highlighting/highlighters/README.md/README.md-highlighter.zsh'._

this is caused by a really strange behaviour of zsh itself.

**ZSH Issue Example**

If you create a file and folder

```
mkdir -p folder
touch file
```

...,if you want to print only folders globing works as expected
```
echo ./*/
>> ./folder/
```

..., however if you run the same command with sudo it will also print the file, as a folder.
thats totally weird.
```
sudo zsh -c 'echo ./*/'
>> ./file/ ./folder/
```

**Solution**

Because thats a zsh issue I adjust the globing expression, now it works as expected.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -373,7 +373,7 @@ _zsh_highlight_load_highlighters()
 
   # Load highlighters from highlighters directory and check they define required functions.
   local highlighter highlighter_dir
-  for highlighter_dir ($1/*/); do
+  for highlighter_dir ($1/*/(/)); do
     highlighter="${highlighter_dir:t}"
     [[ -f "$highlighter_dir${highlighter}-highlighter.zsh" ]] &&
       . "$highlighter_dir${highlighter}-highlighter.zsh"


danielshahaf:
Cross-referencing #506.  More below

> ..., however if you run the same command with sudo it will also print the file, as a folder.
> thats totally weird.
> 
> sudo zsh -c 'echo ./*/'
> >> ./file/ ./folder/
> 

I can't reproduce that.  Please change -c to -fc and try again.  What's the zsh version, distro, and `mount` output for the relevant mountpoint?

> Solution
> 
> Because thats a zsh issue I adjust the globing expression, now it works as expected.

Thanks for the analysis. I'd still like to double-check whether it's really a bug in the `*/` syntax, before I avoid it.

qoomon:
macOS 10.13.4 (17E202)
zsh 5.5.1 (x86_64-apple-darwin17.5.0)

I changed parameter to `-fc` still wrong result with sudp

danielshahaf:
Thanks.  I'm on Linux, so I'll wait for an OS X -using dev to weigh in.

In the meantime, could you please confirm that `zsh --version` and `sudo zsh --version` are the same value?

danielshahaf:
The zsh bug was reported as workers/42891 : http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=42891

qoomon:
i did this :-D

so i test with ubuntu and i can not reproduce it there, however i tried it on macos of my colleges and it seem like its not a problem only on my mashines

qoomon:
Hey, would be great if you could except my pull request, cause I think it will take some time until zsh is fixed :-( 
And may patch is only 3 characters long :-D

phy1729:
I'd like to move to autoloading the highlighters in the future, but this patch is tiny as you point out, so I think it's reasonable to merge it. Could you rebase into a single commit and match the commit message style? (Something like `main: Add glob qualifier to work around zsh/macOS bug` and perhaps link to the mailing list thread as well.)

qoomon:
done

phy1729:
Thanks for the report and fix!

PR #524, opened 2018-08-20
Add instructions about initializing variable

issmirnov:
Many users use modular configuration loaders for zshrc. Adding a note to users explaining how to define ZSH_HIGHLIGHT_HIGHLIGHTERS before sourcing zsh-syntax-highlighting. The typeset is important, since simply setting the variable doesn't have the desired effect.

### DIFF
--- a/docs/highlighters.md
+++ b/docs/highlighters.md
@@ -27,7 +27,11 @@ To activate an highlighter, add it to the `ZSH_HIGHLIGHT_HIGHLIGHTERS` array in
     ZSH_HIGHLIGHT_HIGHLIGHTERS=(main brackets pattern cursor)
 
 By default, `$ZSH_HIGHLIGHT_HIGHLIGHTERS` is unset and only the `main`
-highlighter is active.
+highlighter is active. 
+
+If you would like to set this variable before
+sourcing zsh-syntax-highlighting, initialize it first with: 
+`typeset -ga ZSH_HIGHLIGHT_HIGHLIGHTERS`
 
 
 How to tweak highlighters


danielshahaf:
It should work without an explicit typeset, though `WARN_CREATE_GLOBAL` might trigger.. and that knob is off by default.

danielshahaf:
I'm confused as to why the typeset would be _required_.

issmirnov:
I could be missing something, but when I tried to get this working on my own [dotfiles](https://github.com/issmirnov/dotfiles/tree/master/zsh) (see `zshrc` and `config/zsh-syntax-highlighting.zsh`) the variables were reset to defaults after sourcing this plugin. Here's a sample `zshrc` that reproduces the issue on my machine (`$ZSH_VERSION=5.3`, OSX 10.13.6) - the last echo for `ZSH_HIGHLIGHT_PATTERNS` comes back empty. 

```
export ZSH_HIGHLIGHT_HIGHLIGHTERS=(main brackets pattern cursor)
export ZSH_HIGHLIGHT_PATTERNS=('ls *' 'fg=white,bold,bg=red')

source ~/.dotfiles/zsh/plugins/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh

echo "(1) defined vars (2) sourced plugin"
echo $ZSH_HIGHLIGHT_HIGHLIGHTERS
echo $ZSH_HIGHLIGHT_PATTERNS
```

Compared with this config, where the `ls` is correctly highlighted.

```
typeset -ga ZSH_HIGHLIGHT_HIGHLIGHTERS
typeset -gA ZSH_HIGHLIGHT_PATTERNS
export ZSH_HIGHLIGHT_HIGHLIGHTERS=(main brackets pattern cursor)
export ZSH_HIGHLIGHT_PATTERNS=('ls *' 'fg=white,bold,bg=red')

source ~/.dotfiles/zsh/plugins/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh

echo "(1) defined vars (2) sourced plugin"
echo $ZSH_HIGHLIGHT_HIGHLIGHTERS
echo $ZSH_HIGHLIGHT_PATTERNS

```

Is there a better way to do this? I tried various permutations but this was the only solution I found that worked for me.

danielshahaf:
Arrays can't be exported.  (I'd argue that should have given you a syntax error, actually.)  If you drop the `export` and leave just `foo=(a b c)` in zshrc, that would declare a global array.

issmirnov:
Ah, interesting! Yes, removing the export works just fine. So looks like `ZSH_HIGHLIGHT_HIGHLIGHTERS ` doesn't need to be touched, but the `typeset -gA` for `ZSH_HIGHLIGHT_PATTERNS ` is needed. I'll close this PR and wait for some comments on the other one.

Thank you for looking into this.

PR #525, opened 2018-08-20
Add instructions about initializing variable

issmirnov:
Many users use modular configuration loaders for zshrc. Adding a note to users explaining how to define ZSH_HIGHLIGHT_PATTERNS before sourcing zsh-syntax-highlighting. The typeset is important, since simply setting the variable doesn't have the desired effect.


issmirnov:
See discussion in #524

danielshahaf:
> See discussion in #524

The cases are similar but there's an important distinction: #524 is about an array (`typeset -ga`) and this issue is about an associative array (`typeset -gA`). Just writing `foo=(1 2 3)` implicitly declares `foo` as a _normal_ array; `typeset -A` must be used before assigning to an associative array.

(I don't use the `pattern` highlighter so I hoped one of the other devs would review this patch.)

Issue #528, opened 2018-08-30
main-highlighter.zsh: hang when highlight word matches dead remote mount in pwd


wraeth:
**Brief Description**
In some circumstances, the main highlighter can hang the shell. This can occur when the word being typed matches a directory in $PWD, and that directory is a mounted remote filesystem (NFS in this case) which has since died (mounted but non-responsive). Adding the directory to `X_ZSH_HIGHLIGHT_DIRS_BLACKLIST` has no effect.

**Steps to replicate**

1. Create directory in $PWD with any arbitrary name (eg. `~/myoldhost`)
2. Mount an NFS share to this directory
3. Make the NFS share non-responsive (in my case: break networking on remote host)
4. Attempt to type a command where an argument happens to be the same name (eg. `ssh myoldhost`)

**Expected result**

Command to be successfully entered and ready to execute

**Actual result**

Shell hangs before last character appears

**Details**

This appears to be caused by the highlighter attempting to probe the named directory, in some cases regardless of whether the argument is a directory/path type, while the mount is dead. Because the system still sees it mounted, the probe is queued waiting for a response resulting in the shell hanging waiting for the syscall to return.

phy1729:
Fixed with #529.

PR #529, opened 2018-08-30
main-highlighter.zsh: Probe dirs after blacklist

wraeth:
Move tests on the highlight path until after the blacklist is checked so
that blacklisted paths are not tested. This should prevent hangs on dead
remote filesystems provided the path is blacklisted in
X_ZSH_HIGHLIGHT_DIRS_BLACKLIST.

Bug: https://github.com/zsh-users/zsh-syntax-highlighting/issues/528

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -761,8 +761,6 @@ _zsh_highlight_main_highlighter_check_path()
   REPLY=path
 
   [[ -z $expanded_path ]] && return 1
-  [[ -L $expanded_path ]] && return 0
-  [[ -e $expanded_path ]] && return 0
 
   # Check if this is a blacklisted path
   if [[ $expanded_path[1] == / ]]; then
@@ -776,6 +774,9 @@ _zsh_highlight_main_highlighter_check_path()
     [[ -n "${(M)X_ZSH_HIGHLIGHT_DIRS_BLACKLIST:#$tmp_path}" ]] && return 1
     tmp_path=$tmp_path:h
   done
+
+  [[ -L $expanded_path ]] && return 0
+  [[ -e $expanded_path ]] && return 0
 
   # Search the path in CDPATH
   local cdpath_dir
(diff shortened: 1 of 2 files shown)


danielshahaf:
I'd say that a syscall not returning is a kernel bug, but the change makes sense. Please fix the test suite failures. Adding a test would be good too, see `tests/generate.zsh`.

wraeth:
I just pushed an updated commit that sets the tests to expect 'default' highlighter on matched paths (ie. I fixed the broken ones). My understanding of the tests is limited to a half-hour of reading, poking, and kludgery, so please let me know how they might be improved.

In terms of adding a new test, I'm not sure what more might need to be tested, given that there's already some tests in place. Suggestions welcome.

danielshahaf:
The test expectations are correct _as they stand in master_. Please fix your patch to pass.

danielshahaf:
I meant, add a regression test for `X_ZSH_HIGHLIGHT_DIRS_BLACKLIST`.

danielshahaf:
Sorry! I hadn't noticed the test sets the envvar already. Ignore my previous comment. Apologies.

danielshahaf:
@phy1729 WDYT?

phy1729:
This was an open question still when the feature was pushed (#496) (and why the X_ remains). Since there's a clear use case, I works for me.

`[[ -z $expanded_path ]] && return 1` should perhaps be before the blacklist check. I'm not sure if it's necessary at all, but it's been there since the start of _check_path (b253ce80).

@wraeth two nits on the commit, convention is just the highlighter name before the colon and there's a stray g in highlight in the second paragraph. Thanks for following up on this!

phy1729:
I think keep the `[[ -z` test before the blacklist check for now and we can decide later if it is needed. No reason for it to block merging.

If you wouldn't mind fixing and force pushing I'd appreciate it. Thanks!

wraeth:
Done, and done.

Let me know if there's anything more.

danielshahaf:
LGTM

Two preëxisting comments (not blockers, just drive-by observations):

1. Maybe look into using an associative array for the blacklist

2. Maybe allow `/` to be blacklisted.

phy1729:
Sorry it took so long to get back to this. Thanks for the PR!

wraeth:
Thanks for the merge!

PR #533, opened 2018-09-15
Add check for escaped brackets in levelpos calculation

### LINKED ISSUE #532: Escaped bracket not ignored
Technically a repost of Issue #138 (the second part of the issue was not addressed in #455)

The following snippet incorrectly renders the last bracket as an error:

```zsh
echo [ \] ]
```

cbcoutinho:
Really simple way to fix #532


phy1729:
I'd use single quotes instead of dollar quotes (e.g. `BUFFER='echo (x\])'`), but that's just style.

What I was trying to convey before was instead of adding conditionals in brackets-highlighter.zsh, you could add under line 56 or so
```
\\)
(( SOME_VAR )) && pos++
;;
```
where `SOME_VAR` is a placeholder for a variable a user could set to opt-in to the behavior (or opt-out. I don't use the brackets highlighter, so I don't have a strong opinion on what the default should be.)

danielshahaf:
@cbcoutinho Why close this?

cbcoutinho:
I accidentally borked a rebase/force push, and gh closed it
automatically... Oops.

I'll re-commit the changes and open it up again.

Op ma 1 okt. 2018 17:10 schreef Daniel Shahaf <notifications@github.com>:

PR #543, opened 2018-10-10
Add Markdown Syntax Highlight

Sea-n:


phy1729:
I think because of https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L4433 , you can use zsh as the language specifier which would allow for github to later highlight zsh specific syntax.

If you have the time, would you mind looking around the other docs and standardizing the snippits to all use ```zsh? If not we can just open an issue to complete later.

Sea-n:
@phy1729 Thanks for the suggestion!

I think I have done the work :smiley:

phy1729:
The changes look good. Thank you for looking over all the docs! Sorry for not getting back to you sooner.

If you're still interested, could you rebase into a single commit with a commit message something like `docs: Enable syntax highlighting for code snippits`? If not, I can do those steps while of course keeping attribution to you on the commit.

Sea-n:
@phy1729 Done 🙃

phy1729:
Merged. Thanks for the PR and your patience!

PR #546, opened 2018-10-10
Use shallow clone while installing

user501254:
Takes much less time on a slow connection.
Also, it doesn't seem practical to have complete commit history for installation purposes.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -33,7 +33,7 @@ See also [repology's cross-distro index](https://repology.org/metapackage/zsh-sy
 
 Simply clone this repository and source the script:
 
-    git clone https://github.com/zsh-users/zsh-syntax-highlighting.git
+    git clone --depth 1 https://github.com/zsh-users/zsh-syntax-highlighting.git
     echo "source ${(q-)PWD}/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh" >> ${ZDOTDIR:-$HOME}/.zshrc
 
   Then, enable syntax highlighting in the current interactive shell:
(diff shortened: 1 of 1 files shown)


user501254:
Hi @danielshahaf 
I do agree that many people would know about the  `--depth` option. However, it's not that uncommon to use it. For example, here is a similar change in [golang's](https://go.googlesource.com/go/+/bc1f9d20b40998b87c9a5f56f2b22595de65c2d4%5E%21/#F0) repository. 
With time, the repository history size will only increase and it can make for a really frustrating 5 minutes in areas with patchy network.

danielshahaf:
I'm sorry, but the answer's still no.  This repository is not large (it's about half the size of a release _tarball_ of zsh itself), and on balance I don't think people on 56kbps modems who don't already know about the `--depth` option are a large enough group to merit cluttering the documentation for everyone else.

```
% units -t '2.1 MiB / 5 minutes' 'bits per second' 
58720.256
```

It's entirely standard for installation instructions to have to be customized for the local setup.

Also, people on narrow connections have the option of using z-sy-h release tarballs.  (Now that I mention it, I'm not even sure that `git clone --depth` is designed to minimise network traffic.)

user501254:
The tragedy is that those people are on so called "4G" connections. 😅
I don't believe that the `--depth` option was specifically designed to minimize network traffic [but the way it works uses less data](https://stackoverflow.com/questions/3954852/how-to-complete-a-git-clone-for-a-big-project-on-an-unstable-connection). 
I respect your decision to have less cluttered installation instructions. You may close this.

danielshahaf:
No other opinions, so closing. Thanks for discussion!

PR #550, opened 2018-10-13
main: Distinguish quoted/unquoted command substitutions

### LINKED ISSUE #547: Distinguish command substitutions that are/aren't quoted
It would be useful to have a visual distinction between `: $(foo)` and `: "$(foo)"`, as the former undergoes word splitting and the latter does not.

phy1729:
Closes #547.


highlighters/main/main-highlighter.zsh
```diff
@@ -107,6 +107,9 @@ _zsh_highlight_main_calculate_fallback() {
       dollar-quoted-argument{-unclosed,}
       back-quoted-argument{-unclosed,}
 
+      command-substitution{-quoted,,-unquoted,}
+      command-substitution-delimiter{-quoted,,-unquoted,}
```
danielshahaf: Thanks for implementing this!  My only comment is that the new styles should be added to the documentation.

phy1729: Indeed. Thanks for the reminder. Fixed.

PR #551, opened 2018-10-13
main: Break double-quoted-argument on command substitutions

phy1729:
This breaks the double-quoted-argument highlighting
```zsh
echo "foo$(echo bar)baz"
```
into two parts (`"foo` and `baz"`) which results in the command substitution and it's contents not being highlighted as quoted (unless of course there are quotes within the command substitution).


danielshahaf:
Would it be better to change the `default` style to something like `fg=0`?  The result should be the same then, but one could then still use `underline` or `bg=…` styles that apply to the entire double-quoted string, including the cmdsubst.

highlighters/main/test-data/dollar-paren.zsh
```diff
@@ -28,14 +28,22 @@
 # vim: ft=zsh sw=2 ts=2 et
 # -------------------------------------------------------------------------------------------------
 
-BUFFER=': "$(:)"'
+BUFFER=': "$(:)" "foo$(:)bar'
```
danielshahaf: Worth adding a comment that the string literal is intentionally unterminated.

phy1729:
`fg=0` would result in text being unreadable in a terminal with a black background and `fg=7` would result in the same on a terminal with a white background. The base style (usually default) still applies to the entirety of the `(z)`-word.

danielshahaf:
That, and `command-substitution-quoted` applies just to the `$( …)` part, so it ought to still be possible to customize the rendering.

Go ahead, then?  Not sure if this requires special mention in changelog / release notes, but I'm sure you'll add it if it's needed.

PR #553, opened 2018-10-15
append .zshrc in new line

Ir1d:


Vaelatern:
How does this help?

Ir1d:
closing since regarded as invalid

danielshahaf:
It was marked invalid due to being concomitant with a deluge of Hacktoberfest spam PRs. If we false positived we apologize.

That aside, it's questionable whether this is needed (files generally _do_ have trailing newlines, and readability counts). Furthermore, backslash in double quotes isn't an escape character in all shells (e.g., dash).

Issue #540, opened 2018-10-01
Aliases of precommand with flag that takes an argument


phy1729:
```
alias sdu='sudo -u'
sdu phy1729 echo foo
```

Should highlight phy1729 as normal (as it's an argument) and echo as a command. The parser currently treats phy1729 as the command.

danielshahaf:
I'm surprised `phy1729` is taken to be a command word.

phy1729:
The issue is that even once precommands is merged only the first word of an alias is run through the state machine.

danielshahaf:
Duplicate of newer #552.

PR #556, opened 2018-10-18
Parse full alias

phy1729:
This runs the entirety of the alias through the state machine not just the first (z)-word.

There's still an issue when the non-terminal alias has more than one (z)-word, but I don't have the brain power to think through that tonight.

This changes behavior for aliases. An alias will be highlighted as alias and the style appropriate for the first (z)-word of the fully resolved alias.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -533,20 +533,20 @@ _zsh_highlight_main_highlighter_highlight_list()
       local res="$REPLY"
       if [[ $res == "alias" ]]; then
         () {
-        local -A seen_arg
+        local -A seen_alias
         while [[ $REPLY == alias ]]; do
-        seen_arg[$arg]=1
+        seen_alias[$arg]=1
         _zsh_highlight_main__resolve_alias $arg
         # Use a temporary array to ensure the subscript is interpreted as
         # an array subscript, not as a scalar subscript
-        local -a reply
+        local -a alias_args
         # TODO: the ${interactive_comments+set} path needs to skip comments; see test-data/alias-comment1.zsh
-        reply=( ${interactive_comments-${(z)REPLY}}
-                ${interactive_comments+${(zZ+c+)REPLY}} )
+        alias_args=( ${interactive_comments-${(z)REPLY}}
+                     ${interactive_comments+${(zZ+c+)REPLY}} )
         # Avoid looping forever on alias a=b b=c c=b, but allow alias foo='foo bar'
-        [[ $arg == $reply[1] ]] && break
-        arg=$reply[1]
-        if (( $+seen_arg[$arg] )); then
+        [[ $arg == $alias_args[1] ]] && break
+        arg=$alias_args[1]
+        if (( $+seen_alias[$arg] )); then
           res=none
           break
         fi
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -128,22 +128,36 @@ _zsh_highlight_main_calculate_fallback() {
 # Uses the zsh/parameter module if available to avoid forks, and a
 # wrapper around 'type -w' as fallback.
 #
-# Takes a single argument.
+# If $2 is 0, do not consider aliases.
 #
 # The result will be stored in REPLY.
 _zsh_highlight_main__type() {
+  integer -r aliases_allowed=${2-1}
+  # We won't cache replies of anything that exists as an alias at all, to
+  # ensure the cached value is correct regardless of $aliases_allowed.
+  #
+  # ### We probably _should_ cache them in a cache that's keyed on the value of
+  # ### $aliases_allowed, on the assumption that aliases are the common case.
+  integer may_cache=1
+
+  # Cache lookup
   if (( $+_zsh_highlight_main__command_type_cache )); then
     REPLY=$_zsh_highlight_main__command_type_cache[(e)$1]
     if [[ -n "$REPLY" ]]; then
       return
     fi
   fi
+
+  # Main logic
   if (( $#options_to_set )); then
     setopt localoptions $options_to_set;
   fi
   unset REPLY
   if zmodload -e zsh/parameter; then
     if (( $+aliases[(e)$1] )); then
+      may_cache=0
+    fi
+    if (( $+aliases[(e)$1] )) && (( aliases_allowed )); then
       REPLY=alias
     elif (( $+saliases[(e)${1##*.}] )); then
       REPLY='suffix alias'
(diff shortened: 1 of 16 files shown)


phy1729:
Re-ready for review; incorporated feedback from IRC.

highlighters/main/main-highlighter.zsh
```diff
@@ -128,22 +128,36 @@ _zsh_highlight_main_calculate_fallback() {
...
 # wrapper around 'type -w' as fallback.
 #
-# Takes a single argument.
+# If $2 is 0, do not consider aliases.
 #
 # The result will be stored in REPLY.
 _zsh_highlight_main__type() {
+  integer -r aliases_allowed=${2-1}
+  # We won't cache replies of anything that exists as an
+  # alias at all, regardless of $aliases_allowed.
```
danielshahaf: Say why: because the __type of a word that exists as an alias depends on the value of $aliases_allowed.  (That's why the "We probably should" in the next sentence makes sense.)

highlighters/main/main-highlighter.zsh
```diff
@@ -531,62 +544,46 @@ _zsh_highlight_main_highlighter_highlight_list()
...
+            ;;
           # - The common case:
           (*) :;;
         esac
-        if (( insane_alias )); then
-          style=unknown-token
-        # Calling 'type' again; since __type memoizes the answer, this call is just a hash lookup.
-        elif ! _zsh_highlight_main__type "$arg" || [[ $REPLY == 'none' ]]; then
-          style=unknown-token
+        args=( $alias_args $args )
```
danielshahaf: IRC proposed `"$foo[@]"` for belt and braces.

highlighters/main/main-highlighter.zsh
```diff
@@ -455,7 +468,7 @@ _zsh_highlight_main_highlighter_highlight_list()
       fi
     fi
 
-    if true; then
+    if (( in_alias == 0 )); then
```
danielshahaf: Does this DTRT on `alias foo='x && y' y=ls` followed by `% foo`?

highlighters/main/main-highlighter.zsh
```diff
@@ -531,62 +544,46 @@ _zsh_highlight_main_highlighter_highlight_list()
...
         esac
-        if (( insane_alias )); then
-          style=unknown-token
-        # Calling 'type' again; since __type memoizes the answer, this call is just a hash lookup.
-        elif ! _zsh_highlight_main__type "$arg" || [[ $REPLY == 'none' ]]; then
-          style=unknown-token
+        args=( $alias_args $args )
+        if (( in_alias == 0 )); then
+          _zsh_highlight_main_add_region_highlight $start_pos $end_pos alias
+          (( in_alias += $#alias_args + 1 )) # Add one because in_alias counts $arg as well
```
danielshahaf: If I understand correctly, the ` + 1` is because the _current_ iteration handles the alias, and the next `$#alias_args` iterations will handle its expansion, is that correct?  I think it might be clearer if it were said in more words?

highlighters/main/main-highlighter.zsh
```diff
@@ -531,62 +544,46 @@ _zsh_highlight_main_highlighter_highlight_list()
...
         else
-          # The common case.
-          style=alias
-          if (( ${+precommand_options[(re)"$arg"]} )) && (( ! ${+precommand_options[(re)"$arg_raw"]} )); then
-            precommand_options[$arg_raw]=$precommand_options[$arg]
-          fi
+          (( in_alias += $#alias_args ))
```
danielshahaf: I don't see why this branch doesn't also have a `+ 1`?  The number of words we push to $args does not depend on whether the current word is a result of alias expansion, does it?

highlighters/main/main-highlighter.zsh
```diff
@@ -531,62 +544,46 @@ _zsh_highlight_main_highlighter_highlight_list()
...
         fi
-        }
+        (( in_redirection++ )) # Stall this arg
```
danielshahaf: So now `in_redirection` is used by redirections, by `repeat`, and by aliases.  I'm starting to think we should at least rename it, lest we cause bugs by overly overloading it.

highlighters/main/test-data/alias-precommand-option-argument1.zsh
```diff
@@ -35,7 +35,8 @@ BUFFER='sdu phy1729 echo foo'
 
 expected_region_highlight=(
   '1 3 alias' # sdu
-  '5 11 default "issue #540"' # phy1729
+  '1 3 precommand' # sdu (sudo)
+  '5 11 default' # phy1729
   '13 16 commmand "issue #540"' # echo (not builtin)
   '18 20 default' # foo
```
danielshahaf: Please reference #540 in the log message, as this fixes one form of it.

highlighters/main/main-highlighter.zsh
phy1729: The +1 is because we will `in_alias--` on the next loop iteration so this iteration should be considered in `in_alias` as well. In the case that we're already `in_alias` this arg is already included in the count. It makes sense in my brain; hopefully that helps?

highlighters/main/main-highlighter.zsh
phy1729: I think more interesting would be `y=sudo` and yes.

danielshahaf:
> The +1 is because we will `in_alias--` on the next loop iteration so 
> this iteration should be considered in `in_alias` as well. In the case 
> that we're already `in_alias` this arg is already included in the count. 
> It makes sense in my brain; hopefully that helps?

Yes; add it as a source comment?

highlighters/main/main-highlighter.zsh
phy1729: I think elision is wanted here because `alias x=''` _should_ be elided (and is with the assignment of `alias_args` anyway).

highlighters/main/main-highlighter.zsh
```diff
@@ -72,6 +72,9 @@ _zsh_highlight_main_add_region_highlight() {
   integer start=$1 end=$2
   shift 2
 
+  (( highlighted_alias )) && return
```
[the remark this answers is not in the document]
phy1729: Added some tests for this.

highlighters/main/main-highlighter.zsh
phy1729: Not entirely mechanical since 
```
_zsh_highlight_main_highlighter_highlight_argument 1 $(( 1 - in_redirection ))
```
may need to be changed.

highlighters/main/main-highlighter.zsh
phy1729: Comment updated.

highlighters/main/main-highlighter.zsh
phy1729: Comment added.

highlighters/main/main-highlighter.zsh
danielshahaf: Sure

danielshahaf:
The new test 6898f7101652 fails on this branch.

phy1729:
Since `BUFFER=/` is highlighted as `path`, I'd argue that either the new behavior is correct, or the behavior of `BUFFER=/` is wrong as well and it's a separate issue.

Issue #406, opened 2016-12-29
Highlight case/esac


danielshahaf:
In `case foo in (bar) ls;; esac`:

- [ ] The `ls` not highlighted as a command word.
- [ ] The `esac` is not highlighted as a reserved word.
- [ ] The `case`/`esac` pair should be subject to balanced-parenthesis logic, like `do`/`done`.  (So `esac` is highlighted in red if not preceded by a `case`, etc.)

PR #563, opened 2018-10-21
Fix function definition highlighting

### LINKED ISSUE #223: Highlight function definitions
In `f() {}`, highlight `f` specially.  Ditto for `f g h() {}` when MULTI_FUNC_DEF is in effect (that would help with a common user mistake).

The default color of `f` `g` `h` shall be different from the default color of the [command] style.

The first word is parsed differently from the others: it is subject to alias expansion and to must not be a reserved word.  That said, using a reserved word for `g` or `h` would probably be a mistake on the user's part (so it might be useful to check for that and highlight it as unknown-token).

The first word should never be considered for the [command] and [path] highlights.

Further ideas (which might be good and might be pure feature creep):
- [ ] Highlight the syntax `function f g h () {}` as well?
- [ ] Highlight `f` differently depending on whether it shadows/overwrites an existing function, builtin, or command?

Compare #194.

phy1729:
Fixes #223 

Adds new `function-definition` style; thoughts on what that should be by default?


docs/highlighters/main.md
```diff
@@ -55,6 +55,7 @@ This highlighter defines the following styles:
 * `dollar-double-quoted-argument` - parameter expansion inside double quotes (`$foo` inside `""`)
 * `back-double-quoted-argument` -  backslash escape sequences inside double-quoted arguments (`\"` in `"foo\"bar"`)
 * `back-dollar-quoted-argument` -  backslash escape sequences inside dollar-quoted arguments (`\x` in `$'\x48'`)
+* `function-definition` - function name(s) when defining a function (`f` and `g` in `f g () { : }`)
```
danielshahaf: Test cases (written before I read the diff):

`function { pwd }` — anonymous function

`function pwd` — defines a function named `pwd`, _not_ runs that command

`function >/dev/null { pwd }` — anonymous function

`function f >/dev/null { pwd }` — named function with sticky redirection

`function >/dev/null f { pwd }` — syntax error (!)

The above also support an optional `()` after the function names and before the brace.

`() {}` — anonymous function

`() pwd` — anonymous function

`() () () pwd` — anonymous function for calling tor

The above also come in named variants.

`function () () pwd` — two anonymous functions

`function f () () pwd` — named function whose body is an anonymous function

`function () f () pwd` — anonymous function that defines a function named `f`

`function () () f pwd` — two anonymous functions, `f` is in command position (unknown-token)

Then there's the MULTI_FUNC_DEF variants of all these.

Then there's

`unsetopt multifuncdef; function f g h { pwd }` — creates a function named `f`.  Probably a zsh bug, actually, but shouldn't we highlight `g` and `h` as errors nevertheless?  (I saw the test for the non-`function`-syntax variant of this.)


Question: In `() pwd`, should the `()` be highlighted as a command position?  What about `function { pwd }`?

highlighters/main/main-highlighter.zsh
```diff
@@ -751,6 +765,10 @@ _zsh_highlight_main_highlighter_highlight_list()
                            #
                            # The repeat-count word will be handled like a redirection target.
                            this_word=':start::regular:'
+                           ;;
+                        ('function')
+                           next_word=:function:
```
danielshahaf: Minor: I usually quote `':foo:'` tokens, partly for Vim syntax highlighting purposes.  We should probably be consistent one way or the other.

highlighters/main/main-highlighter.zsh
```diff
@@ -846,16 +864,9 @@ _zsh_highlight_main_highlighter_highlight_list()
...
-                 else
-                   if [[ $zsyh_user_options[multifuncdef] == on ]] || false # TODO: or if the previous word was a command word
-                   then
-                     next_word+=':start:'
-                   fi
-                   style=reserved-word
-                 fi
+        '()')
+                 # Function definition was handled above
+                 style=unknown-token
```
danielshahaf: When can this happen?  `f g ()` with nomultifuncdef, `function "foo'bar" ()`, …?

highlighters/main/main-highlighter.zsh
```diff
@@ -677,8 +678,21 @@ _zsh_highlight_main_highlighter_highlight_list()
...
+       next_word=:start:
+     else
+       style=function-definition
+       next_word=:function:
+     fi
    elif ! (( in_redirection)) && [[ $this_word == *':start:'* ]]; then # $arg is the command word
-     if (( ${+precommand_options[$arg]} )) && _zsh_highlight_main__is_runnable $arg; then
+     if [[ $res != reserved && ( $args[1] == '()' ||
+           # TODO: Function names can be absurd; this handles the more common cases without invoking Cthulhu.
+           ( $zsyh_user_options[multifuncdef] == on && $args[(r)*[^[:alnum:]_-]*] == '()' ) ) ]]; then
```
danielshahaf: I wonder if we could do something along the line of "after `function`, everything that isn't a token (`&&`, `>`, etc) or the reserved words `{` or `()` is a function name?  Haven't looked at Src/parse.c of this.

docs/highlighters/main.md
danielshahaf: Another case (github didn't let me edit comment after I pressed the green button):

```zsh
% function f
pwd
% f
```

highlighters/main/main-highlighter.zsh
phy1729: or after a redirection or in an array assignment.

highlighters/main/main-highlighter.zsh
phy1729: `function` already has that in effect. This branch only effects words before `()`.

highlighters/main/main-highlighter.zsh
danielshahaf: What about empty `case` patterns?  `case foo in (bar) ;; esac` → `case foo in () ;; esac`

highlighters/main/main-highlighter.zsh
danielshahaf: Sorry, that's what I meant. "If one of the following words is `()`, and it's in the same simple command as the current word, then all intervening words [other than redirections] are function names" or something like this.  Basically, we know that "everything from the start of the command to the `()` token is function names", with some exceptions (`noglob`, redirections, …), so I wonder if a blacklist might be better than a whitelist.

highlighters/main/main-highlighter.zsh
phy1729: Ah, I was wondering why many were quoted. I'll match.

highlighters/main/main-highlighter.zsh
phy1729: Right now it's highlighted as `reserved-word` which isn't correct either. We'll probably need another state to handle case correctly.

danielshahaf: With this branch the `()` in `case` is highlighted as unknown-token — which is correct, since it's a syntax error. Let's discuss `case` on #406.

docs/highlighters/main.md
danielshahaf: How about highlighting it the same way as the newlines in 

```zsh
for 1 in foo bar baz
do
  echo $1
done
```

?

danielshahaf:
@phy1729 This is milestoned 0.8.0. Do you intend to work on it for the 0.8.0 release? If yes, great; if not, could you re-milestone it appropriately? Cheers.  If it's waiting on me, let me know!

danielshahaf:
(Assigning until the question is answered.  Feel free to unassign yourself if needed.)

danielshahaf:
Triage: Bumping to the next milestone due to feedback timeout.

PR #566, opened 2018-10-22
'main': Highlight the ! precommand as an error when not at the start of a pipeline.

### LINKED ISSUE #511: Incorrectly placed ! not highlighted as syntax error
In `echo | ! cat` `!` is highlighted as `reserved-word` but should be highlighted as `unknown-token` as it's a syntax error (since `!` can only occur at the start of a pipeline).

danielshahaf:
Fixes #511.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -743,17 +743,17 @@ _zsh_highlight_main_highlighter_highlight_list()
                           ('end')
                             _zsh_highlight_main__stack_pop '$' reserved-word
                             ;;
-                         ('repeat')
-                           # skip the repeat-count word
-                           in_redirection=2
-                           # The redirection mechanism assumes $this_word describes the word
-                           # following the redirection.  Make it so.
-                           #
-                           # That word can be a command word with shortloops (`repeat 2 ls`)
-                           # or a command separator (`repeat 2; ls` or `repeat 2; do ls; done`).
-                           #
-                           # The repeat-count word will be handled like a redirection target.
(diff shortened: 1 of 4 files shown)


highlighters/main/main-highlighter.zsh
```diff
@@ -822,15 +833,15 @@ _zsh_highlight_main_highlighter_highlight_list()
...
+      next_word=':start::start_of_pipeline:'
      fi
    else # $arg is a non-command word
       case $arg in
         $'\x29') # subshell or end of array assignment
                  if $in_array_assignment; then
                    style=assign
                    in_array_assignment=false
-                   next_word+=':start:'
+                   next_word+=':start::start_of_pipeline:'
```
phy1729: `foo=(bar) ! false` is a parse error here, so I think this is incorrect.

highlighters/main/main-highlighter.zsh
```diff
@@ -684,6 +689,9 @@ _zsh_highlight_main_highlighter_highlight_list()
       case $res in
         reserved)       # reserved word
                         style=reserved-word
+                        if [[ $arg == '!' && $this_word != *':start_of_pipeline:'* ]]; then
```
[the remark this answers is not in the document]
danielshahaf: Changed.

highlighters/main/main-highlighter.zsh
danielshahaf: Fixed.

danielshahaf:
Added tests and addressed the comments.  Thanks for the review.

PR #590, opened 2018-12-12
Add instruction to declare the variable

gnumoksha:


danielshahaf:
I've extended the fix to the regexp highlighter, tweaked the log message and merged. Thanks for the patch!

PR #593, opened 2018-12-19
Create various globals with typeset -g

okdana:
I decided recently that i'd turn on `warn_create_global`, but a lot of my `zshrc` stuff happens inside functions to avoid polluting my session with temp variables, and z-sy-h doesn't play entirely nice with that:

```
% zsh --warn-create-global -fc '() { source ./zsh-syntax-highlighting.zsh }'
./zsh-syntax-highlighting.zsh:31: scalar parameter zsh_highlight__aliases created globally in function (anon)
./zsh-syntax-highlighting.zsh:60: scalar parameter zsh_highlight__pat_static_bug created globally in function (anon)
```

Not sure if there are other places where this is an issue, but just changing those two seems to sort it for me.

Edit: Fixed the failing tests on old zsh, rebased

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -28,7 +28,7 @@
 # -------------------------------------------------------------------------------------------------
 
 # First of all, ensure predictable parsing.
-zsh_highlight__aliases=`builtin alias -Lm '[^+]*'`
+typeset zsh_highlight__aliases="$(builtin alias -Lm '[^+]*')"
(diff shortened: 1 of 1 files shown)


danielshahaf:
Why -g? Shouldn't plain `typeset`  suffice?

okdana:
Sorry, i didn't look at what they were actually doing.

It seems like the first one is safe to make local (and i have done, rebased again) — but the second one is used by `_zsh_highlight`, so i think that would cause problems here. Let me know if i'm mistaken though

okdana:
Like so?

danielshahaf:
Perfect. Will merge when at my workstation. Thanks!

PR #599, opened 2019-01-07
Highlight non-matching globs as errors or normal

phy1729:
Obviously needs tests, doc updates, and `files=($~arg(NY1))` is fragile, but what do you think of the idea? Basically 
```
: /nonexistant?
```
Will highlight `?` as `unknown-token` if no_match is set and not highlight at all otherwise.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -978,7 +978,7 @@ _zsh_highlight_main_highlighter_check_path()
 _zsh_highlight_main_highlighter_highlight_argument()
 {
   local base_style=default i=$1 option_eligible=${2:-1} path_eligible=1 ret start style
-  local -a highlights
+  local -a globs highlights
 
   local -a match mbegin mend
   local MATCH; integer MBEGIN MEND
@@ -1087,6 +1087,7 @@ _zsh_highlight_main_highlighter_highlight_argument()
           highlights+=($(( start_pos + i - 1 )) $(( start_pos + i + $#MATCH - 1)) globbing)
           (( i += $#MATCH - 1 ))
           path_eligible=0
+          globs+=($#highlights)
         else
           continue
         fi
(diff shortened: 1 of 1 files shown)


highlighters/main/main-highlighter.zsh
```diff
@@ -1098,6 +1099,20 @@ _zsh_highlight_main_highlighter_highlight_argument()
     base_style=$REPLY
     _zsh_highlight_main_highlighter_highlight_path_separators $base_style
     highlights+=($reply)
+  elif (( $#globs )); then
+    local glob
+    local -a files; files=($~arg(NY1))
```
danielshahaf: `(Y1)` is new in 5.x.  In versions that don't have it, any `**` in there could be very expensive to compute (and of course it'll be recomputed every `self-insert`).

`$~foo` performs filename generation which can eval arbitrary code via `*(+foo)`.  We specifically avoid it in `_zsh_highlight_main_highlighter_expand_path`.

highlighters/main/main-highlighter.zsh
```diff
@@ -1098,6 +1099,20 @@ _zsh_highlight_main_highlighter_highlight_argument()
...
+    if ! [[ -e $files[1] ]]; then
```
danielshahaf: Minor issue, but we could avoi the `stat` by simply checking `$#files` (it'll 0 or 1 so no need for the faster `$+files[1]`).

highlighters/main/main-highlighter.zsh
phy1729: Yep. The globs we match are all safe, but we don't parse any of the unsafe globs. This feature can't happen until all unsafe globs are recognized. Sorry, must have been tired.

danielshahaf: Not sure I follow. We shouldn't blacklist dangerous globs, but whitelist known-good ones.

phy1729: While I agree whitelisting is better, we would first have to recognize all globs in order to only allow whitelisted ones. Or we could whitelist safe globs and non-globbing characters and assume anything else is dangerous.

Issue #592, opened 2018-12-18
sudo and commands only in /sbin


Raizo62:
Hi

Sorry, i come back for an other problem with sudo

With "sudo ls" , "ls" is colorized correctly with green

with "sudo ifup" where "ifup" is only in "/sbin", "ifup" is colorized in red

Raizo62:
Can z-sy-h define a specific variable used only with "sudo" code of z-sy-h which contains default path of root's commands (as /sbin).
If an user defines an alternative path for root in sudoers, then he can modify this variable ?

danielshahaf:
> Can z-sy-h define a specific variable used only with "sudo" code of 
> z-sy-h which contains default path of root's commands (as /sbin).
> If an user defines an alternative path for root in sudoers, then he can 
> modify this variable ?

Such users would also need to send the command-path style in the
:completion:*:sudo:* zstyle context.  It would be nice to have
completion and z-sy-h both obtain the value from the same place.

danielshahaf:
`zstyle ":completion:*:sudo:*" command-path /bin /sbin ...`. Please ask further completion questions on zsh-users@ or #zsh on FreeNode to distribute the workload.

/sbin _is_ in users' PATH on FreeBSD.

Good luck with your course!

Raizo62:
> `zstyle ":completion:*:sudo:*" command-path /bin /sbin ...`. Please ask further completion questions on zsh-users@ or #zsh on FreeNode to distribute the workload.

Sorry, i believe that i was not been clear : I didn't ask how to do the completion on zsh for /sbin.
I said that the completion of zsh can have different actions with the word "sudo" .
When i am user, the completion of zsh doesn't show "ifup" , but if i use the word "sudo", zsh shows "ifup"
Then my question was : why can z-sy-h not do the same ?
 
> /sbin _is_ in users' PATH on FreeBSD.

It is not in Debian, and i think that is a good thing ;-)

danielshahaf:
Because that hasn't been implemented. Patches welcome.

Raizo62:
I can try....

danielshahaf:
Great! Let's open a new issue to discuss possible approaches.

Raizo62:
I think to add in "_zsh_highlight_main__type" something as 
`if $(sudo which "$1" >/dev/null) ; then; REPLY=command; fi`
[...]

PR #610, opened 2019-03-31
Patch to color more correctly the command behind "sudo"

Raizo62:
if the command is `sudo foo` with : 
- ZSH_HIGHLIGHT_SPECIAL_PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
if `foo` is in the PATH ${ZSH_HIGHLIGHT_SPECIAL_PATH} then it is green
- ZSH_HIGHLIGHT_SPECIAL_PATH=""
`foo` is red
- ZSH_HIGHLIGHT_SPECIAL_COMMAND=(brctl ls)
Only if `foo` is `brctl` or `ls` then `foo` is green ; other are red
- (nothing) :  
default behavior

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -127,6 +127,18 @@ _zsh_highlight_main_calculate_fallback() {
   done
 }
 
+_zsh_highlight_main__test_type_builtin() {
+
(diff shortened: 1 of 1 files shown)


danielshahaf:
Discussed on #592.

@Raizo62 I'll review later.

PR #622, opened 2019-07-07
driver: ksharrays issue with $options

austintraver:
I personally use ksh-style arrays, but I noticed that this plugin doesn't work when that setting is enabled. Adding this one line of code allows this plugin to work for users who include "setopt ksharrays" in their .zshrc file.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -85,6 +85,8 @@ _zsh_highlight()
 
   # Before we 'emulate -L', save the user's options
   local -A zsyh_user_options
+  # Retain 1-indexing of zsh arrays
+  setopt localoptions noksharrays
   if zmodload -e zsh/parameter; then
     zsyh_user_options=("${(@kv)options}")
   else
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -91,7 +91,7 @@ _zsh_highlight()
     local canonical_options onoff option raw_options
     raw_options=(${(f)"$(emulate -R zsh; set -o)"})
     canonical_options=(${${${(M)raw_options:#*off}%% *}#no} ${${(M)raw_options:#*on}%% *})
-    for option in $canonical_options; do
+    for option in $("${(kv)options[@]}"); do
       [[ -o $option ]]
       # This variable cannot be eliminated c.f. workers/42101.
       onoff=${${=:-off on}[2-$?]}
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -85,10 +85,8 @@ _zsh_highlight()
 
   # Before we 'emulate -L', save the user's options
   local -A zsyh_user_options
-  # Retain 1-indexing of zsh arrays
-  setopt localoptions noksharrays
   if zmodload -e zsh/parameter; then
-    zsyh_user_options=("${(@kv)options}")
+    zsyh_user_options=("${(kv)options[@]}")
   else
     local canonical_options onoff option raw_options
     raw_options=(${(f)"$(emulate -R zsh; set -o)"})


danielshahaf:
I don't think that's quite the right fix: it'll cause `zsyh_user_options` to be initialized with wrong values.  Furthermore, the `emulate -L zsh` on line 103 already resets KSH_ARRAYS to the default value (off).

Presumably got a syntax error somewhere between line 88 (where you added the unsetopt) and line 102 (where the unsetopt already happens as part of `emulate -L`).

Could you propose a different fix?

danielshahaf:
By the way, the travis errors are due to #623; not your fault.  Sorry about that.

phy1729:
This appears to be an upstream bug.
```zsh
% PROMPT='%% ' zsh -f
% print $#options
192
% setopt ksharrays
% print $#options 
3
% print $options
off
```
In addition, forcing the fallback path results in canonical_options being set to just `(aliases)` even though raw_options looks to be correct. (Of course `onoff=${${=:-off on}[2-$?]}` needs to be adjusted when running under ksharrays as well.)

danielshahaf:
This works:

```zsh
% Src/zsh -fc 'setopt ksharrays; echo ${#${options[@]}}' 
192
% 
```

As to ${#options}: given that `Src/zsh -fc 'setopt ksharrays; echo ${options}'` prints `off`, I suspect ${#options} computes the length of the scalar `"off"`.  I don't know whether that's a bug or an expected effect of KSH_ARRAYS.

phy1729:
Didn't think about `[@]`. If that works, this may suffice (when zsh/parameter is available).
```diff
diff --git a/zsh-syntax-highlighting.zsh b/zsh-syntax-highlighting.zsh
index df8a1aa..36ab6e9 100644
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -86,7 +86,7 @@ _zsh_highlight()
   # Before we 'emulate -L', save the user's options
   local -A zsyh_user_options
   if zmodload -e zsh/parameter; then
-    zsyh_user_options=("${(@kv)options}")
+    zsyh_user_options=("${(kv)options[@]}")
   else
     local canonical_options onoff option raw_options
     raw_options=(${(f)"$(emulate -R zsh; set -o)"})
```

danielshahaf:
Fixed the travis issue; the next push to this PR should be green.

austintraver:
Any other changes you guys would like to see before this gets pulled?

danielshahaf:
Could you fix the `else` branch as well?

Could you add comments (or a regression test)?

zsh-syntax-highlighting.zsh
```diff
@@ -86,12 +86,12 @@ _zsh_highlight()
...
   local -A zsyh_user_options
   if zmodload -e zsh/parameter; then
-    zsyh_user_options=("${(@kv)options}")
+    zsyh_user_options=("${(kv)options[@]}")
   else
     local canonical_options onoff option raw_options
     raw_options=(${(f)"$(emulate -R zsh; set -o)"})
     canonical_options=(${${${(M)raw_options:#*off}%% *}#no} ${${(M)raw_options:#*on}%% *})
-    for option in $canonical_options; do
+    for option in $("${(kv)options[@]}"); do
```
danielshahaf: This is incorrect: `$options` doesn't exist in the `else` branch.

chitoku-k:
@austintraver (cc: @danielshahaf)
I found this issue can be affected by other plugins even when they just locally enable `ksharrays`. Do you happen to have time to fix them?

danielshahaf:
@chitoku-k I don't have time to fix this personally.  If anyone has patches to propose, though, I'm all ears.  For starters, can you check if merging just the first of the two changes in this PR makes a difference?

austintraver:
> @austintraver (cc: @danielshahaf)
> I found this issue can be affected by other plugins even when they just locally enable `ksharrays`. Do you happen to have time to fix them?

For me, this was a case of wanting a configuration set before I really realized what I was missing out on. After fiddling with the repo, and successfully making a version of this script that could handle `ksharrays`,  eventually it broke again, and I just decided to throw my hands up and use the default array style.

I like it much better actually, so I've switched sides. I'm afraid my previous solution no longer works, it was broken a while back.

But that didn't feel like a helpful thing to say, so I ended up writing a solution tonight. Closing this pull request in favor of the continuation I wrote in https://github.com/zsh-users/zsh-syntax-highlighting/pull/689

Issue #637, opened 2019-10-13
Characters are highlighted misleadingly in vi-mode visual selection


doronbehar:
I'm experiencing an annoyance using `bindkey -v` and visual mode selection only when this plugin is enabled. I'll explain:

Say my $BUFFER has the following text, current $KEYMAP is "vicmd", and
the cursor is located where ^ is:

```
ZSH is the best shell
              ^
```

I press `vF<space>` in order to highlight the characters marked with `-----`:

```
ZSH is the best shell
          ^
          -----            (selected characters)
```
Depending on the contents of the `$BUFFER`, sometimes the letter `t` of the word `best` isn't highlighted, for example:

If my `$BUFFER` has:

```
git commit -m "ZSH is the best shell"
```

The string `ZSH is the best shell` will be colored (`fg=yellow` with my config) and performing the same actions will result in the letter `t` not highlighted.

In comparison, if the region selected isn't affected by z-sy-h, or the plugin is completely disabled, the letter `t` **is highlighted**.

danielshahaf:
Works for me, xterm 327-2, xfce4-terminal 0.8.3-1, zsh HEAD, z-sy-h HEAD.

Could you please show the output of `typeset -p ZSH_VERSION ZSH_PATCHLEVEL ZSH_HIGHLIGHT_REVISION ZSH_HIGHLIGHT_VERSION`? What's the name and version of your terminal emulator?

And humour me, but could we please use [neutral text](https://en.wikipedia.org/wiki/Lorem_ipsum) for bug reports?

danielshahaf:
Can you reproduce this in `zsh -f` + `source zsh-syntax-highlighting.zsh`?

doronbehar:
Daniel here's a demonstration of the behavior using a desktop recording:

![z-sy-h-nohighlight-issue](https://user-images.githubusercontent.com/10998835/66747939-5966bb80-ee75-11e9-931e-4c278ce3537f.gif)

As you can see, the letter `t` isn't highlighted when the cursor is at the other end of the selection, only when this part of the buffer is already handled by z-sy-h. Otherwise, `t` _is_ highlighted.

I'm using urxvt and here's my [`.Xresources`](https://github.com/doronbehar/dotfiles/blob/5d9d8ca/.Xresources#L103).

Here's my urxvt version:

```.
rxvt-unicode (urxvt) v9.22 - released: 2016-01-23
[...]

PR #638, opened 2019-10-17
adjust region end for vi command mode

okapia:
This is a fix for issue #637.

You apparently need to replicate the logic from zle_refresh.c to redraw the region. This needs a slight fixup for vi command mode where in the C, we have:
```
        } else if (invicmdmode())
            INCPOS(region_highlights[0].end);
```
The shell code equivalent is essentially:
```
    [[ $KEYMAP = vicmd ]] && (( max++ ))
```
This adjustment roughly corresponds to the cursor adjustment you get when entering vi command mode. The change is slightly bigger because _zsh_highlight_apply_zle_highlight arguments don't have a fixed order so we need to swap them to get the max. You may want to do things differently, such as in that function directly.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -157,24 +157,25 @@ _zsh_highlight()
     # Re-apply zle_highlight settings
 
     # region
-    if (( REGION_ACTIVE == 1 )); then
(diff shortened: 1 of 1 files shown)


danielshahaf:
Thank you very much!

PR #653, opened 2019-12-27
Fix highlighting of recursive aliases

danielshahaf:
See log messages for details.

@NonLogicalDev Please review/test.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -547,7 +547,7 @@ _zsh_highlight_main_highlighter_highlight_list()
 
     if [[ $this_word == *:start:* ]] && ! (( in_redirection )); then
       # Expand aliases.
-      _zsh_highlight_main__type "$arg"
+      _zsh_highlight_main__type "$arg" "$(( ! ${+seen_alias[$arg]} ))"
       local res="$REPLY"
       if [[ $res == "alias" ]] && [[ $last_alias != $arg ]]; then
         # Avoid looping forever on alias a=b b=c c=b, but allow alias foo='foo bar'
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -393,9 +393,7 @@ _zsh_highlight_main_highlighter_highlight_list()
   # alias_style is the style to apply to an alias once in_alias=0
   #     Usually 'alias' but set to 'unknown-token' if any word expanded from
   #     the alias would be highlighted as unknown-token
-  # last_alias is the last alias arg (lhs) expanded (if in an alias).
-  #     This allows for expanding alias ls='ls -l' while avoiding loops.
-  local alias_style arg buf=$4 highlight_glob=true last_alias style
+  local alias_style arg buf=$4 highlight_glob=true style
   local in_array_assignment=false # true between 'a=(' and the matching ')'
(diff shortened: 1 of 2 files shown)


NonLogicalDev:
@danielshahaf This has fixed the issue for me 👍 ! Thank you so much!

NonLogicalDev:
The more I look at this:
```
alias rm=rmi
alias rmi='nocorrect rm -i'
```

The more I am amazed and confused about how zsh is parsing this... Logic should dictate that this should lead to an endless loop, but yet zsh somehow handles this just fine...

NonLogicalDev:
I found more edge cases:

Edge 01:
```
alias te=st
```
Leads to `te` being unknown (Perhaps Intended)

Edge 02:
```
alias te=nocorrect st
```
Leads to `te` colored as alias (Perhaps Unintended??)

danielshahaf:
> The more I look at this the more I am amazed and confused about how zsh is parsing this... One would assume that this should lead to an endless loop, but yet zsh somehow handles this just fine...

Yeah, even `alias ls=ls` works, even though `alias a=b b=ls` proves that RHS of aliases are subject to alias expansion.

There's also a rule that if the RHS of an alias ends with a space, the _next_ word is subject to alias expansion... which means things like `alias nocorrect="nocorrect " ls="t " t="nocorrect ls "` should also be loops — but they work.

> Edit: Interestingly enough here `rm` works but `rmi` fails. So I would assume that this is actually not a bug, but a correct highlighting.

Yes, `rmi` fails so highlighting it as an error (= in red) is correct.

> ```
> alias te=st
> ```
> 
> Leads to `te` being unknown (Perhaps Intended)

It's intended if you don't have a command/function/alias called `st` (or a directory called `st` and autocd is set).

> ```
> alias te=nocorrect st
> ```
> 
> Leads to `te` colored as alias (Perhaps Unintended)

No, it's correct.  You create an alias `te` whose value is `nocorrect`, which is valid in command position so `te` is highlighted in green; and then you asked zsh to show you the value of the alias `st`, which presumably doesn't exist, so nothing was printed and `alias` exited non-zero.  (See the printexitvalue option and the %? prompt expando.)  If you had had an `st` alias, it'd have been printed.

There's an argument to be made that zsh should print `alias: st not found` or something to that effect, but that's a question for zsh-users@zsh.org.

danielshahaf:
> Yeah, even `alias ls=ls` works, even though `alias a=b b=ls` proves that RHS of aliases are subject to alias expansion.

I've checked the source, and it turns out that [an alias is ineligible for expansion while it's being expanded](https://github.com/zsh-users/zsh/blob/4a2a15d8dc8d310d537ababc2abba0fe444276a8/Src/lex.c#L1826-L1828); i.e., circular definitions are not possible. That's why `alias nosuchcommand=nosuchcommand` just reports that `nosuchcommand` is not a valid command name.

Thanks for all your testing! :)

NonLogicalDev:
Thank you for getting on this so quickly, and explanations, and even pointing to some source of ZSH, perhaps I will take a dive into it someday.

phy1729:
Makes sense to me. I think the

> Avoid looping forever on alias a=b b=c c=b, but allow alias foo='foo bar'

comment can go as well with the removal of `(( $+seen_alias[$arg] ))` (or be moved to the __type call if you prefer).

danielshahaf:
Will update comments and merge. Thanks for the review.

Issue #658, opened 2020-01-11
precommand flags with an unknown argument


danielshahaf:
I think `sudo -x ls` should _not_ be highlighted as an error, for forward compatibility with future versions of sudo.  Therefore, I think the test added in e27e53b1ef19387949f4a214c641c568539bb027 is wrong.

@phy1729 WDYT?

(FWIW, there was actually a comment about this very situation in the code before that commit — not that that matters now.)

phy1729:
It's impossible for z-sy-h to know aprori if `-x` takes an argument. While it's true we currently assume it doesn't, at least there's an indication of where the parsing went wrong.

If there were a style for known syntax error distinct from unknown-token, unknown-token could be defaulted to normal masking such issues for end users.

danielshahaf:
There are several different matters here.

First, was there a particular use-case or instance that prompted you to make the change?  If so, could you describe it, for context?

> It's impossible for z-sy-h to know aprori if -x takes an argument. While it's true we currently assume it doesn't,

Agreed: the code should make that assumption.  It should only assume the normal BSD option stacking conventions (as in `_arguments -s : "-◌+${colon_or_null}"`; we don't generally  know whether the `-◌` option takes an argument).  Therefore, `sudo -x foo bar` should highlight `-x` as `single-hyphen-option` and _both_ of the following words as `default` — exactly the same as `ls -x foo bar`, except for the command word.

That's neither what the code before the referenced commit did nor what HEAD does.

> at least there's an indication of where the parsing went wrong.

We shouldn't use unknown-token unless we know positively that there's a syntax error; showing bold red on valid command lines would be poor UX.  (Call it the Hippocratic Principle if you wish, or observe that ${EDITOR}s generally behave this way.)

If we've lost track of the state, we should fall back to `default` until we can resync (e.g., by encountering a global alias or a command separator).

Information about where parsing went wrong is like debug printf's: of interest to the developers but not to users.  Users just want their tools to stay out of their (the users') ways so they (the users) will be able to get on with whatever they're doing.

> If there were a style for known syntax error distinct from unknown-token, unknown-token could be defaulted to normal masking such issues for end users.

It is not a "known" syntax error.  `${precommand_options}` isn't the source of truth here.  It _could_ be a typo on the part of the user, but it could also be a new upstream release of `sudo` that the code doesn't know about.  Don't confuse the dependent and independent variables :-) 

[...]

PR #659, opened 2020-01-11
'main': precommands += ssh-agent

danielshahaf:
This touches on the matters being discussed in #658, so not pushing it directly.  If unknown flags are to be highlighted as unknown-token, `k` should be added to the RHS; if they're to be highlighted as single-hyphen-option, it should be left out.

I suspect the specific case of `ssh-agent -k`, which may not be followed by further arguments, falls under "Hard cases make bad law" insofar as the design question in #658 is concerned.


danielshahaf:
Thanks for the review.  They are in the same order as the man page's synopsis, but not in the same order as the alphabetical list of flags further down the man page.

phy1729:
Ah, ok.

Issue #641, opened 2019-11-09
sudo long flags not recognised


baltitenger:
Long flags for sudo (e.g. `--askpass` or `--preserve-env`) are treated as commands, and highlighted for not existing. The plugin also seems to assume `sudo [flags] [env] [command] [args...]` order, but flags and env can be mixed. (e.g. in `sudo -s foo=bar baz`, it recogdies `-s` as a flag, `foo=bar` as an environment variable, and `baz` as a command. but in `sudo foo=bar -s baz`, `foo=bar` is recognised as env, `-s` as a command, and `baz` as an argument)

danielshahaf:
This is a multi-pronged issue; let me break it down.

> Long flags for sudo (e.g. `--askpass` or `--preserve-env`) are treated as commands, and highlighted for not existing.

- [ ] 641.1 `--foo` flags that don't take an arguments: Easy.
- [ ] 641.2 `--foo=bar` flags that take an argument in the same word: Easy.
- [ ] 641.3 `--foo bar` flags that take an argument in the next word: Would require hardcoding the list of such options, as we already do for one-letter options.
- [ ] 641.4 `--` end-of-options marker: Easy.

> The plugin also seems to assume `sudo [flags] [env] [command] [args...]` order, but flags and env can be mixed.

> (e.g. in `sudo -s foo=bar baz`, it recogdies `-s` as a flag, `foo=bar` as an environment variable, and `baz` as a command.

This is a bug.  The assignment is highlighted here for the same reason it's highlighted in `% LC_ALL=C ls`, but the code is generic and most precommands _don't_ allow assignments here; `sudo` and `env` are exceptions.  So:

- [ ] 641.5 `nice x=y ls`: Should be `unknown-token`.  Easy to support with a little state machine work.
- [ ] 641.6 `sudo -s foo=bar ls`, `env x=y ls`: Currently work.  Are exceptions to the case in the previous bullet.

> but in `sudo foo=bar -s baz`, `foo=bar` is recognised as env, `-s` as a command, and `baz` as an argument)

- [ ] 641.7 `sudo foo=bar -s ls`: Could be supported, but it's unique to `sudo`.

danielshahaf:
Tentatively setting milestone to 0.8.0, but it's not a release blocker.

danielshahaf:
[...]

PR #664, opened 2020-01-12
Fix unknown precommand flags being highlighted as unknown-token

### LINKED ISSUE #658: precommand flags with an unknown argument

danielshahaf:
Fixes #658, see also #641.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -664,12 +664,9 @@ _zsh_highlight_main_highlighter_highlight_list()
           next_word+=':start:'
           next_word+=':sudo_opt:'
         elif [[ $arg == '-'* ]]; then
-          # Unknown flag
+          # Unknown flag.  We don't know whether it takes an argument or not,
+          # so we don't modify $next_word.
           this_word=':sudo_opt:'
-          next_word+=':start:'
-          next_word+=':sudo_opt:'
-          _zsh_highlight_main_add_region_highlight $start_pos $end_pos unknown-token
-          continue
         else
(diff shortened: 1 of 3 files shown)


highlighters/main/main-highlighter.zsh
```diff
@@ -664,12 +664,9 @@ _zsh_highlight_main_highlighter_highlight_list()
...
         elif [[ $arg == '-'* ]]; then
-          # Unknown flag
+          # Unknown flag.  We don't know whether it takes an argument or not,
+          # so we don't modify $next_word.
           this_word=':sudo_opt:'
-          next_word+=':start:'
-          next_word+=':sudo_opt:'
-          _zsh_highlight_main_add_region_highlight $start_pos $end_pos unknown-token
-          continue
         else
```
danielshahaf: Discussed this on IRC with @phy1729. Our takeaway was to re-add the `next_word+=` settings.  That will fix #641.0 and #658's issue about unknown flags being red, and restore 0.6.0's behaviour of assuming unknown flags take no argument.  With that behaviour, flags that take no argument will highlight the inner command word correctly, and flags that take an argument will highlight the command word correctly if the argument is given in the same shell word as the flag (as in `-uphy1729` or `--user=phy1729` without spaces).  That's compared to the branch's current behaviour of never highlighting the inner command word at all.

Issue #670, opened 2020-01-12
Assignment as parameter value


danielshahaf:
First case: (#670.1)

```zsh
% foobar='x=$(ls)'
% $foobar
```

highlights the `ba` in green from `ls`, as foreseen in https://github.com/zsh-users/zsh-syntax-highlighting/issues/572#issuecomment-485036695.

Second case: (#670.2)

Exactly the same with a shorter parameter name:

```zsh
% y='x=$(ls)'
% $y
zsh-syntax-highlighting: BUG: _zsh_highlight_highlighter_main_paint: start(2) >= end(2)
```

danielshahaf:
Any relation to the following? —

https://github.com/zsh-users/zsh-syntax-highlighting/blob/521fedfdf5facbe0f45e331253c924758624e864/highlighters/main/test-data/alias-command-substitution.zsh#L30-L37

danielshahaf:
Added to changelog as a known issue.

Analysis:

`$arg` is assigned to but `start_pos`/`end_pos` are unchanged, so highlighting is usually correct.  The highlighting would be incorrect when `$arg` expands to something that equals a token, a reserved word, etc..  That's what happens with assignments: the `=` is taken to be syntactically significant even though it's not.  That causes the code that recognizes assignments in command position to kick in, and that code assumes that `$arg[i]` is okay to be highlighted.

danielshahaf:
@phy1729 Have you made progress on reviewing #675, https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:parameters-multi-word-machinery-v2, or the kludge I proposed on IRC (also pasted below)?

```diff
From 7c04246ec8e94e3e340dd4c0c063695e873593af Mon Sep 17 00:00:00 2001
From: Daniel Shahaf <danielsh@apache.org>
Date: Sat, 25 Jan 2020 21:43:18 +0000
Subject: [PATCH 1/2] 'main': Add a kludge for issue #670, concerning command
 position parameter expansions that expand to things that look like
 assignments, other tokens, or reseved words.
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

If the tests added in the next (child) commit are run on the previous
(parent) commit, they fail as follows:

    % make quiet-test
    make test QUIET=y
    make[1]: Entering directory '/home/daniel/in/zsh-syntax-highlighting'
[...]

PR #675, opened 2020-01-16
WIP: Parse multi-word parameters in command position

### LINKED ISSUE #674: Parse array parameters in command position
- [x] Add array parameter versions of the alias-precommand-option-argument*.zsh tests
- [ ] Make them pass (cf. #670) - pass array parameters through the alias machinery

danielshahaf:
Fixes #674. Improves the behaviour for #670: there's no longer a `BUG:` message on stderr, but the tests don't pass (the error is recorded in the log message).  Conflicts with #667.

The code is based on the aliases code.  Some edge cases are probably missing.

In particular, I think what's happening is that in aliases `alias_style` is initialized to `alias` but in parameters, `param_style` isn't analogously initialized, so that's why the `assign` style never gets set.  The `return` in `_zsh_highlight_main_add_region_highlight` makes `assign` not be recorded anywhere.

Hints welcome.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -77,6 +77,10 @@ _zsh_highlight_main_add_region_highlight() {
     [[ $1 == unknown-token ]] && alias_style=unknown-token
     return
   fi
+  if (( in_param )); then
+    [[ $1 == unknown-token ]] && param_style=unknown-token
+    return
+  fi
 
   # The calculation was relative to $buf but region_highlight is relative to $BUFFER.
   (( start += buf_offset ))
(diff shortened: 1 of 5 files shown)


danielshahaf:
Pushed a few more commits.  Now it passes tests except that it adds the style twice in the two new tests for #670.  I still don't know if the patch is correct or not, though…

danielshahaf:
I've pushed a revised version of the branch to [danielsh/parameters-multi-word-machinery-v2](https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:parameters-multi-word-machinery-v2?expand=1).  The changes are:

- Squash the oldest and third-oldest commits, so now there's three commits (rather than four) and the oldest of the remaining ones passes tests.
- Extend the log message of the second commit.

danielshahaf:
> I've pushed a revised version of the branch to [danielsh/parameters-multi-word-machinery-v2](https://github.com/zsh-users/zsh-syntax-highlighting/compare/master...danielshahaf:parameters-multi-word-machinery-v2?expand=1). The changes are:

That's #684, now merged.  It supersedes this PR.

PR #696, opened 2020-03-15
Fix highlighting of multiline aliases

### LINKED ISSUE #677: alias including a newline is not highlighted anymore
With the config
```zsh
alias foobar='echo foo;
  echo bar'
```
`foobar` at the head of the prompt should be highlighted like a command but not highlighted. This used to work before the commit 9cc0060.

danielshahaf:
Exempt such aliases from the empty commands sanity check.

Does anyone see a downside to making this change?  The argument is basically that it's not z-sy-h's job to lint the user's dotfiles, but I wonder if there's a counter-argument that I'm overlooking.

Fixes #677, which knowingly regressed in 0.7.0 compared to 0.6.0..

This didn't require converting `';'` back to `$'\n'` in alias expansions after all.

@itchyny Please test/review.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -85,6 +85,10 @@
   `_zsh_highlight_main__precmd_hook:1: array parameter _zsh_highlight_main__command_type_cache set in enclosing scope in function _zsh_highlight_main__precmd_hook`
   [#727, #731, #732, #733]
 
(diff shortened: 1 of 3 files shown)


danielshahaf:
Rebased due to b454b596edc661813bee4eef24282a5aa729eaa0.

Issue #265, opened 2016-01-14
Highlight global aliases


danielshahaf:
Idea: Highlight global aliases, created by `alias -g`.

danielshahaf:
Would you be interested in writing the patch? We'd be happy to point you in the right direction.

HaleTom:
I'd be happy to scope out the work and see if it's something that fits my expertise.

How to be in touch?

danielshahaf:
Thanks. The bases to cover are the implementation, documentation, and a unit test; see `grep -R 'suffix.alias'` and mimic it. You can talk to us here or on #zsh-syntax-highlighting on freenode. (I'm not always online, but I'll read scrollback and reply.)

HaleTom:
If anyone else wishes to merge in the changes, this has been implemented by @psprint here: https://github.com/zdharma/fast-syntax-highlighting/issues/41

danielshahaf:
@HaleTom Thanks for the reference, but it doesn't get us any closer to implementing the same feature in z-sy-h.  We just need someone to write the patch and PR it.  (It doesn't have to be you, of course.)

If you could give examples of "faster" and "more complex syntax", that'd be appreciated.  There are some known [performance issues](https://github.com/zsh-users/zsh-syntax-highlighting/issues?q=is%3Aopen+label%3Aperformance), but I don't remember any particular outstanding feature requests for smarter parsing.  Cheers!

HaleTom:
The project page details it better than I could.

danielshahaf:
Submitted #700 to fix this.  For reference, incorporating https://github.com/zdharma/fast-syntax-highlighting/'s features is issue #434.

PR #699, opened 2020-03-17
Fix historical instances of one-space indentation.

danielshahaf:
No functional change.

Use-case: I was looking into #265 and ran into the boundary between the even-number-of-spaces and odd-number-of-spaces indentation, so I figured I'd fix this once and for all.

Review with `git show -w --word-diff`.

@phy1729 Would this cause conflicts to any outstanding PR of yours?


danielshahaf:
> Use-case: I was looking into #265 and ran into the boundary between the even-number-of-spaces and odd-number-of-spaces indentation, so I figured I'd fix this once and for all.

See #700.

PR #702, opened 2020-03-17
Highlight escaped reserved words properly

### LINKED ISSUE #701: Escaped variable-declaring reserved words are builtins
In `BUFFER='\local a=( /* )'`, the word `local` is highlighted as a reserved word where it's in fact a builtin.

danielshahaf:
Fixes #701.

There's a TODO in the last commit.


highlighters/main/main-highlighter.zsh
```diff
@@ -153,19 +153,17 @@ _zsh_highlight_main_calculate_fallback() {
 #
 # If $2 is 0, do not consider aliases.
 #
+# If $3 is 0, do not consider reserved words.
+#
```
danielshahaf: A call to this function was added to master after this branch was crested. See whether it should be updated.

PR #703, opened 2020-03-17
'main': Highlight errors from the EQUALS option.

### LINKED ISSUE #430: Highlight EQUALS command name filename expansion as error
`: =foo` should highlight `=foo` as an error ([unknown-token]) when

1. The EQUALS option is set
2. The expansion would fail, as in

        % : =foo
        zsh: foo not found

danielshahaf:
Fixes #430.

### DIFF
--- a/highlighters/main/test-data/equals1.zsh
+++ b/highlighters/main/test-data/equals1.zsh
@@ -1,0 +1,36 @@
+#!/usr/bin/env zsh
+# -------------------------------------------------------------------------------------------------
+# Copyright (c) 2020 zsh-syntax-highlighting contributors
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are permitted
+# provided that the following conditions are met:
+#
+#  * Redistributions of source code must retain the above copyright notice, this list of conditions
+#    and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright notice, this list of
+#    conditions and the following disclaimer in the documentation and/or other materials provided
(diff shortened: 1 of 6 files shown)


highlighters/main/test-data/equals1.zsh
```diff
@@ -0,0 +1,37 @@
...
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+# IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
+# OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# -------------------------------------------------------------------------------------------------
+# -*- mode: zsh; sh-indentation: 2; indent-tabs-mode: nil; sh-basic-offset: 2; -*-
+# vim: ft=zsh sw=2 ts=2 et
+# -------------------------------------------------------------------------------------------------
+
+# TODO: write a complementary test, for BUFFER=': =nosuchcommand' with 'unsetopt equals' in effect
```
danielshahaf: Do this; it probably needed e6eea1f9b7e67136911536a2450c7130f44c0b04.

PR #709, opened 2020-03-19
Support the arithmetic for loop

danielshahaf:
@phy1729 This could benefit from the arithmetic expressions highlighter you wrote yesterday :)


highlighters/main/main-highlighter.zsh
```diff
@@ -1033,6 +1043,9 @@ _zsh_highlight_main_highlighter_highlight_list()
       fi
     elif _zsh_highlight_main__type "$arg"; [[ $REPLY == 'global alias' ]]; then # $arg is a global alias that isn't in command position
       style=global-alias
+    elif [[ $this_word == *':arithmetic_for:'* && $arg == $'\x28\x28' ]]; then # arithmetic for loop
+      braces_stack='4'"$braces_stack"
+      style=reserved-word
```
danielshahaf, line 1048: See if we can use #714 here, as it was just merged.

PR #714, opened 2020-03-22
Arithmetic expansion

phy1729:
Needs docs and a changlog update (and perhaps a few more tests if any come to mind). One outstanding question is should `"$(( ... ))"` break the double quoted style as `"$( ... )"` does?

### DIFF
--- a/docs/highlighters/main.md
+++ b/docs/highlighters/main.md
@@ -42,6 +42,7 @@ This highlighter defines the following styles:
 * `command-substitution-delimiter-quoted` - a quoted command substitution delimiters (`"$(` and `)"`)
 * `process-substitution` - process substitutions (`<(echo foo)`)
 * `process-substitution-delimiter` - process substitution delimiters (`<(` and `)`)
+* `arithmetic-expansion` - arithmetic expansion `$(( 42 ))`)
 * `single-hyphen-option` - single-hyphen options (`-o`)
 * `double-hyphen-option` - double-hyphen options (`--option`)
 * `back-quoted-argument` - backtick command substitution (`` `foo` ``)
--- a/changelog.md
+++ b/changelog.md
@@ -50,13 +50,6 @@
 - Fix `echo >&p` highlighting the `p` as a filename if a file by that name happened to exist
   [part of #645]
 
-- Fix `: $((42))` being highlighted as a subshell.
-  [part of #607]
-
-- Regress highlighting of `: $((ls); (ls))`: is a subshell, but will now be
-  incorrectly highlighted as an arithmetic expansion.
-  [#704]
-
 - Fix wrong highlighting of unquoted parameter expansions under zsh 5.2 and older
   [e165f18c758e]
 
@@ -89,6 +82,8 @@
   (such as `;`, `|`, `&&`) before a newline
   [#677; had regressed in 0.7.0]
 
+- Highlight arithmetic expansions (e.g., `$(( 42 ))`)
+  [#607 #649 #704]
 
 # Changes in version 0.7.1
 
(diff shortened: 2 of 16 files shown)


phy1729:
Added docs, changelog entry, and some more tests. Ready for review.

Outstanding questions:
- Should "$(( ... ))" break the double quoted style as "$( ... )" does?
- Which style should command substitution inside arithmetic substitution use: -quoted, -unquoted, or no suffix?

docs/highlighters/main.md
```diff
@@ -42,6 +42,7 @@ This highlighter defines the following styles:
 * `command-substitution-delimiter-quoted` - a quoted command substitution delimiters (`"$(` and `)"`)
 * `process-substitution` - process substitutions (`<(echo foo)`)
 * `process-substitution-delimiter` - process substitution delimiters (`<(` and `)`)
+* `arithmetic-expansion` - arithmentic expansion `$(( 42 ))`)
```
danielshahaf: Misspelling: "arithmentic"

danielshahaf:
> Needs docs and a changlog update (and perhaps a few more tests if any come to mind).

Thinking out loud:

- Things that look like commands, but aren't:

    - [ ] `$((telnet * 42))` (multiplication)
    - [ ] `$((ls /etc))` (division)
    - [ ] `$((diff - foo))` (subtraction)
    - [ ] `$((zsh +1 -f /etc/passwd))` (addition, subtraction, and division)
    - [ ] `$((ls & ls))` (bitwise and)

- Things that look like arithmetic, but aren't:

    - [ ] `$((ls ls))` (syntax error)
    - [ ] `$(() function ())` (anonymous function)
    - [ ] `$((echo case foo in bar\)); echo ';;')` (subshell)
    - [ ] `$((echo case foo in bar\))` (prefix of previous)
    - [ ] `$((: ")) )(((" ))` (what do you _expect_ this to do, without running it?  Arithmetic, subshell, or syntax error?)
    - [ ] `$((: ") ))(((" ))` (same question)

(I'm not saying that all of these be added; I'm just brainstorming.)

The rule is [in lex.c](https://github.com/zsh-users/zsh/blob/aa93901db2ce33724cc2bbeeaaae0c1e53bd02b3/Src/lex.c#L483-L496).

> Should "$(( ... ))" break the double quoted style as "$( ... )" does?

The criteria I would consider are:

- Consistency across different types of quoting.

- Forward compatibility with highlighting other forms of arithmetic expressions: e.g., subscripts, `repeat` loop counts, arithmetic `for` loops' control expressions. 

- Which approach would give theme authors more options.  (E.g., if the possibilities with approach A are a strict superset of the possibilities with approach B, that'd be an argument in favour of the former.)

> Which style should command substitution inside arithmetic substitution use: -quoted, -unquoted, or no suffix?

What kind of processing (word-splitting, etc) will it undergo?  If it's the same processing in `"$(foo)"` and in this case, then it could use the same highlighting.

The same goes for `$foo`, by the way.  In `$(($foo + $bar))`, both variables are double-quoted, right?  So it'd be nice to use the double-quoted variables style (never mind that we don't have have a style for non-double-quoted variables).

> Ready for review.

Ship it :-)

danielshahaf:
Triage: Ready to merge, and there's still enough time prior to tagging to find any issues: let's merge it!  @phy1729, any reason not to?  If not, would you do the honours?

danielshahaf:
Noting here that https://github.com/zsh-users/zsh-syntax-highlighting/pull/709, once merged, would benefit from integration to this.  (Though it's not clear yet which PR will be merged first.)

phy1729:
Thanks for the reminder. Merged.

PR #727, opened 2020-05-03
Avoid warning under setopt warn_nested_var

erydo:
Previously in a shell running with `setopt warn_nested_var`, the
following warning would be produced:
> ```
> _zsh_highlight_main__precmd_hook:1: array parameter _zsh_highlight_main__command_type_cache set in enclosing scope in function _zsh_highlight_main__precmd_hook
> ```

The documentation for `WARN_NESTED_VAR` is:
> Print a warning message when an existing parameter from an enclosing function scope, or  global, is set  in a function by an assignment or in  math context. […] Where possible, use of `typeset -g` to set the parameter suppresses the error, but note that this needs to be used every time the parameter is  set.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -1677,7 +1677,7 @@ _zsh_highlight_main_highlighter_expand_path()
 # -------------------------------------------------------------------------------------------------
 
 _zsh_highlight_main__precmd_hook() {
-  _zsh_highlight_main__command_type_cache=()
+  typeset -gA _zsh_highlight_main__command_type_cache=()
 }
 
 autoload -Uz add-zsh-hook


erydo:
The CI failure appears unrelated to this PR's changes:
```
/work/highlighters/main/main-highlighter.zsh:1680: parse error near `}'
```

danielshahaf:
No, it's a genuine failure.  (Note it's on the correct line number and didn't happen on `master`.)  Older zsh's don't support the reserved word version of `typeset`, only the builtin version, so they take the empty parentheses as an empty set of glob qualifiers.  Does it work as two lines? — `typeset -gA _zsh_highlight_main__command_type_cache; _zsh_highlight_main__command_type_cache=()`

Once redrawhook is merged we should also merge #688 to put an end to this class of problem.

erydo:
Blast, you're right! Unfortunately it does not suppress the warning if done as two lines. I figured this was a quick fix but it looks like this would be a problem for zsh < 5.1. There are probably workarounds but I am not very familiar with the older versions. Should I close this PR and just leave an issue about the warning?

danielshahaf:
> Unfortunately it does not suppress the warning if done as two lines.

Ah, my bad.  I missed that the documentation actually warns about this.

> There are probably workarounds but I am not very familiar with the older versions.

The main thing is that you can't use `=(…)` in the same line as `typeset` (or one of its aliases, e.g., `local`).  We also avoid relying on special parameters provided by the zsh/parameter module: for example, we do use `${options}`, but we have a fallback codepath for when zsh/parameter isn't available.

Anyway, I just pushed a different fix.  Thanks for the report.

Issue #746, opened 2020-07-01
Improvement in highlighters/regexp/README.md


jnooree:
In macOS, the word boundary syntax "\b" [is not supported](https://stackoverflow.com/questions/9792702/does-bash-support-word-boundary-regular-expressions).

I think it's a good idea to add comment for macOS users, to use "[[:<:]]" and "[[:>:]]" rather than "\b".

Thank you for reading.

danielshahaf:
The stackoverflow question you referenced is entitled _Does bash support word boundary regular expressions?_ and z-sy-h does not support bash.  What does `() { typeset -pm ZSH\*; for 1; [[ $1 =~ '\bfoo' ]] && echo $1 } foo barfoo` output on your system?

jnooree:
Hello, thanks for responding. This is the output.
```
typeset -g ZSH_AUTOSUGGEST_ORIGINAL_WIDGET_PREFIX=autosuggest-orig-
typeset -g ZSH_THEME_GIT_PROMPT_CLEAN=$'%{\C-[[34m%})'
typeset -g ZSH_AUTOSUGGEST_COMPLETIONS_PTY_NAME=zsh_autosuggest_completion_pty
typeset -g ZSH_HIGHLIGHT_VERSION=0.7.2-dev
typeset -g ZSH_AUTOSUGGEST_HIGHLIGHT_STYLE='fg=8'
typeset -g -a ZSH_AUTOSUGGEST_CLEAR_WIDGETS=( history-search-forward history-search-backward history-beginning-search-forward history-beginning-search-backward history-substring-search-up history-substring-search-down up-line-or-beginning-search down-line-or-beginning-search up-line-or-history down-line-or-history accept-line copy-earlier-word )
typeset -g ZSH_VERSION=5.7.1
typeset -g ZSH_PATCHLEVEL=zsh-5.7.1-0-g8b89d0d
typeset -g -A ZSH_HIGHLIGHT_PATTERNS=( )
typeset -g -a ZSH_AUTOSUGGEST_ACCEPT_WIDGETS=( forward-char end-of-line vi-forward-char vi-end-of-line vi-add-eol )
typeset -g ZSH_THEME_RVM_PROMPT_OPTIONS='i v g'
typeset -g -a ZSH_HIGHLIGHT_HIGHLIGHTERS=( main brackets regexp root )
typeset -g ZSH_THEME_TERM_TITLE_IDLE=%n@%m
[...]

Issue #722, opened 2020-04-09
Release 0.8.0?


danielshahaf:
I wonder if we should release master as 0.8.0 so we can immediately after that merge redrawhook, which is overdue. Thoughts? How long should the pre-release freeze be (waiting for bug reports from users of HEAD)?

Does current master perform acceptably?

----

EDIT: Here's a list of some tasks; interdependent tasks are separated by rules:

----

- [x] Check performance — Checked and filed #728.  Further discussion should happen there.

----

- [x] #579 Fix z-asug interoperability: https://github.com/zsh-users/zsh-autosuggestions/issues/529
- [x] Merge feature/redrawhook: #245 #749
    - [x] Triage [widget-wrapping](https://github.com/zsh-users/zsh-syntax-highlighting/labels/widget-wrapping) issues
    - [x] Triage [feature:redrawhook](https://github.com/zsh-users/zsh-syntax-highlighting/labels/feature%3Aredrawhook) issues
- [ ] (nice to have) Merge https://github.com/zsh-users/zsh-syntax-highlighting/issues/688#issuecomment-599104145

----

- [x] Send https://github.com/zsh-users/zsh/pull/57 to zsh upstream — done.
- [x] Merge our patch https://github.com/zsh-users/zsh-syntax-highlighting/issues/418#issuecomment-623054593 that relies on it (fixes https://github.com/okapia/zsh-viexchange/issues/1#issuecomment-623054765) — ~~postponed to 0.9.0 since it can't be released before the zsh patch is merged _and released_~~ — will be in 0.8.0 after all, see https://github.com/zsh-users/zsh-syntax-highlighting/issues/418#issuecomment-650142375.
- [ ] Review [the changelog entry for](https://github.com/zsh-users/zsh-syntax-highlighting/blob/master/changelog.md#notice-about-an-improbable-but-not-impossible-forward-incompatibility) #418 ~~(it's right at the top of changelog)~~

----

- [x] Merge #747

----

- [ ] (nice to have) Merge #706

phy1729:
I had assumed redrawhook was getting merged this cycle. If memory serves it's ready for merge just wasn't because it was late in the 0.7.0 cycle. I can take another look tonight if you like,

danielshahaf:
Could do it that way too. It's just there's a bunch of changes already in for days so I wondered whether to split the release cycle so as to get them out before destabilizing master again. Cheers

danielshahaf:
[...]

PR #747, opened 2020-07-01
Add comments for the platform dependency of the regex module

jnooree:
In BSD-based platform such as macOS, the metacharacters with leading "\\" are not supported.
This adds comments for users with those OS, to avoid confusing results.

Also changes the wrong associative array name in the example(s).

### DIFF
--- a/docs/highlighters/regexp.md
+++ b/docs/highlighters/regexp.md
@@ -15,12 +15,13 @@ typeset -A ZSH_HIGHLIGHT_REGEXP
 ZSH_HIGHLIGHT_REGEXP+=('^rm .*' fg="red",bold)
 ```
 
-This will highlight the whole line starting with `rm` (for all operating systems, 
-in contrast to the below example).
+This will highlight the whole line starting with `rm` command (for all
+operating systems, in contrast to the below example).
 
-Some regex patterns are [subject to the host platform][MAN_ZSH_REGEX], especially
-the kernel. To highlight `sudo` only as a complete word, i.e., `sudo cmd`, but 
-not `sudoedit`:
+Some regular expressions are [subject to the host platform][MAN_ZSH_REGEX], 
+especially the kernel. To highlight `sudo` only as a complete word, i.e., 
+`sudo cmd`, but not `sudoedit`, the respective regular expressions for the host
+systems would be:
 
 * GNU-Linux
 
--- a/docs/highlighters/regexp.md
+++ b/docs/highlighters/regexp.md
@@ -10,37 +10,35 @@ patterns.
 To use this highlighter, associate regular expressions with styles in the
 `ZSH_HIGHLIGHT_REGEXP` associative array, for example in `~/.zshrc`:
 
-```zsh
-typeset -A ZSH_HIGHLIGHT_REGEXP
-ZSH_HIGHLIGHT_REGEXP+=('\bsudo\b' fg=123,bold)
-```
+* GNU-Linux
 
-This will highlight "sudo" only as a complete word, i.e., "sudo cmd", but not
-"sudoedit"
+  ```zsh
+  typeset -A ZSH_HIGHLIGHT_REGEXP
+  ZSH_HIGHLIGHT_REGEXP+=('\<sudo\>' fg=123,bold)
+  ```
+
+* BSD-based platforms
+
+  ```zsh
+  typeset -A ZSH_HIGHLIGHT_REGEXP
+  ZSH_HIGHLIGHT_REGEXP+=('[[:<:]]sudo[[:>:]]' fg=123,bold)
+  ```
+
+This will highlight "sudo" only as a complete word, i.e., "sudo cmd", but not "sudoedit".
+
+As in the example, some regex patterns are platform-dependent. Refer to the 
+manual page of `re_format` before applying regular expressions to avoid confusing 
+results. If portability matters, using only [POSIX ERE (extended regular 
+expressions)][POSIX_ERE] could be an option.
 
 The syntax for values is the same as the syntax of "types of highlighting" of
 the zsh builtin `$zle_highlight` array, which is documented in [the `zshzle(1)`
 manual page][zshzle-Character-Highlighting].
 
 See also: [regular expressions tutorial][perlretut], zsh regexp operator `=~`
-in [the `zshmisc(1)` manual page][zshmisc-Conditional-Expressions]
+in [the `zshmisc(1)` manual page][zshmisc-Conditional-Expressions], GNU
 
+[POSIX_ERE]: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04
 [zshzle-Character-Highlighting]: http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Character-Highlighting
 [perlretut]: http://perldoc.perl.org/perlretut.html
 [zshmisc-Conditional-Expressions]: http://zsh.sourceforge.net/Doc/Release/Conditional-Expressions.html#Conditional-Expressions
-
-### Cautions to BSD-based platform users
-
-In BSD-based platform such as macOS, the regex metacharacters with leading backslash (`\`)
-such as `\b`, `\w`, etc. are [not supported](https://stackoverflow.com/a/12696899).
-
-So if you want something like the above example, you should use `[[:<:]]`(matches the beginning of a word)
-with `[[:>:]]`(matches the end of a word). The complete example would be like:
-
-```zsh
-typeset -A ZSH_HIGHLIGHT_REGEXP
-ZSH_HIGHLIGHT_REGEXP+=('[[:<:]]sudo[[:>:]]' fg=123,bold)
-```
-
-For other metacharacters, consider using
-[POSIX ERE](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04) instead.
--- a/docs/highlighters/regexp.md
+++ b/docs/highlighters/regexp.md
@@ -23,7 +23,7 @@ especially the `libc` module. To highlight `sudo` only as a complete word, i.e.,
 `sudo cmd`, but not `sudoedit`, the respective regular expressions for the host
 systems would be:
 
-* Platforms with GNU `libc` (e.g., many GNU/Linux)
+* Platforms with GNU `libc` (e.g., many GNU/Linux distributions)
 
   ```zsh
   typeset -A ZSH_HIGHLIGHT_REGEXP
--- a/docs/highlighters/regexp.md
+++ b/docs/highlighters/regexp.md
@@ -23,7 +23,7 @@ especially the `libc` module. To highlight `sudo` only as a complete word, i.e.,
 `sudo cmd`, but not `sudoedit`, the respective regular expressions for the host
 systems would be:
 
-* Platform with GNU libc (e.g., many GNU/Linux)
+* Platforms with GNU `libc` (e.g., many GNU/Linux)
 
   ```zsh
   typeset -A ZSH_HIGHLIGHT_REGEXP
--- a/docs/highlighters/regexp.md
+++ b/docs/highlighters/regexp.md
@@ -12,25 +12,25 @@ To use this highlighter, associate regular expressions with styles in the
 
 ```zsh
 typeset -A ZSH_HIGHLIGHT_REGEXP
-ZSH_HIGHLIGHT_REGEXP+=('^rm .*' fg="red",bold)
+ZSH_HIGHLIGHT_REGEXP+=('^rm .*' fg=red,bold)
 ```
 
 This will highlight the whole line starting with `rm` command (for all
 operating systems, in contrast to the below example).
 
 Some regular expressions are [subject to the host platform][MAN_ZSH_REGEX], 
-especially the kernel. To highlight `sudo` only as a complete word, i.e., 
+especially the `libc` module. To highlight `sudo` only as a complete word, i.e., 
 `sudo cmd`, but not `sudoedit`, the respective regular expressions for the host
 systems would be:
 
-* GNU-Linux
+* Platform with GNU libc (e.g., many GNU/Linux)
 
   ```zsh
   typeset -A ZSH_HIGHLIGHT_REGEXP
   ZSH_HIGHLIGHT_REGEXP+=('\<sudo\>' fg=123,bold)
   ```
 
-* BSD-based platforms
+* BSD-based platforms (e.g., macOS)
 
   ```zsh
   typeset -A ZSH_HIGHLIGHT_REGEXP
--- a/docs/highlighters/regexp.md
+++ b/docs/highlighters/regexp.md
@@ -10,6 +10,18 @@ patterns.
 To use this highlighter, associate regular expressions with styles in the
 `ZSH_HIGHLIGHT_REGEXP` associative array, for example in `~/.zshrc`:
 
+```zsh
+typeset -A ZSH_HIGHLIGHT_REGEXP
+ZSH_HIGHLIGHT_REGEXP+=('^rm .*' fg="red",bold)
+```
+
+This will highlight the whole line starting with `rm` (for all operating systems, 
+in contrast to the below example).
+
+Some regex patterns are [subject to the host platform][MAN_ZSH_REGEX], especially
+the kernel. To highlight `sudo` only as a complete word, i.e., `sudo cmd`, but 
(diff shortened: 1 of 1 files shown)


docs/highlighters/regexp.md
```diff
@@ -11,7 +11,7 @@ To use this highlighter, associate regular expressions with styles in the
 `ZSH_HIGHLIGHT_REGEXP` associative array, for example in `~/.zshrc`:
 
 ```zsh
-typeset -A ZSH_HIGHLIGHT_PATTERNS
+typeset -A ZSH_HIGHLIGHT_REGEXP
```
danielshahaf, line 14: Committed this part in fb929edc3019.  Thanks!

docs/highlighters/regexp.md
```diff
@@ -28,3 +28,19 @@ in [the `zshmisc(1)` manual page][zshmisc-Conditional-Expressions]
 [zshzle-Character-Highlighting]: http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Character-Highlighting
 [perlretut]: http://perldoc.perl.org/perlretut.html
 [zshmisc-Conditional-Expressions]: http://zsh.sourceforge.net/Doc/Release/Conditional-Expressions.html#Conditional-Expressions
+
+### Cautions to BSD-based platform users
```
danielshahaf, line 32: As written, this could make "BSD-based platform" users feel like second-class citizens, because the example in the manual doesn't work for them.  Adding a section that documents how to adapt the example in the actual documentation to one's platform is less than ideal: see https://m.xkcd.com/1343/.

I would prefer to give a single example that works everywhere, or failing that, to give several examples next to each other, putting all supported platforms on equal footing.  Makes sense?

docs/highlighters/regexp.md
```diff
@@ -28,3 +28,19 @@ in [the `zshmisc(1)` manual page][zshmisc-Conditional-Expressions]
...
+
+In BSD-based platform such as macOS, the regex metacharacters with leading backslash (`\`)
+such as `\b`, `\w`, etc. are [not supported](https://stackoverflow.com/a/12696899).
```
danielshahaf, line 35: We needn't document the precise differences between platforms here; that would be liable to get out of date.  Simply stating that there are platform differences and pointing to the syntax documentation would suffice, IMO.  WDYT?

docs/highlighters/regexp.md
```diff
@@ -28,3 +28,19 @@ in [the `zshmisc(1)` manual page][zshmisc-Conditional-Expressions]
...
+So if you want something like the above example, you should use `[[:<:]]`(matches the beginning of a word)
+with `[[:>:]]`(matches the end of a word). The complete example would be like:
+
+```zsh
+typeset -A ZSH_HIGHLIGHT_REGEXP
+ZSH_HIGHLIGHT_REGEXP+=('[[:<:]]sudo[[:>:]]' fg=123,bold)
+```
+
+For other metacharacters, consider using
+[POSIX ERE](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04) instead.
```
danielshahaf, line 46: Please expand acronyms on first use.

danielshahaf:
By the way, please add the GitHub magic syntax `Fixes #746` to the log message.

jnooree:
I’ve missed some points.
It would be better if I make another fork, or keep working on this fork and make another PR?
I’m not that familiar with open source communities, sorry about that.

danielshahaf:
> It would be better if I make another fork, or keep working on this fork and make another PR?

Neither.  Keep the same fork, and keep the same PR (branch), and push additional commits to it.  I'll squash/split as needed before merging.

> I’m not that familiar with open source communities, sorry about that.

Not to worry; we were all beginners once.  Thanks for the patch!

jnooree:
Just added some commits. Hope this will help.

> It would be better if I make another fork, or keep working on this fork and make another PR?
> Neither. Keep the same fork, and keep the same PR (branch), and push additional commits to it. I'll squash/split as needed before merging.
> I’m not that familiar with open source communities, sorry about that.
> Not to worry; we were all beginners once. Thanks for the patch!

danielshahaf:
Thanks for the revisions.

Regarding the method of making the amendments — by pushing new commits to the existing PR's branch — it was exactly right.

Regarding the content, I like the current example because it serves two purposes — demonstrating how to do something the `pattern` highlighter can't, and demonstrating platform-dependent aspects — but I still wonder whether we could have two examples: first, a simple example that works everywhere, and second, the current, word boundaries example.  I think this change would make the documentation more accessible to first-time readers; i.e., it would flatten the learning curve.  WDYT?

Another question.  The regexp semantics aren't part of z-sy-h; they're a zsh feature that z-sy-h uses.  Therefore, should a note about the OS-dependence of regexp syntax and a reference to the relevant OS-dependent manual page (re_format(7)?) be added to zsh's manual?  The note could go in [the documentation of the `zsh/regex` module](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fregex-Module), for example.  The zsh manual sources are in zsh's repository: https://github.com/zsh-users/zsh/tree/master/Doc/Zsh

No pun intended.

jnooree:
> Regarding the content, I like the current example because it serves two purposes — demonstrating how to do something the `pattern` highlighter can't, and demonstrating platform-dependent aspects — but I still wonder whether we could have two examples: first, a simple example that works everywhere, and second, the current, word boundaries example. I think this change would make the documentation more accessible to first-time readers; i.e., it would flatten the learning curve. WDYT?

That's a good idea. I would add one in front of the current examples.

> Another question. The regexp semantics aren't part of z-sy-h; they're a zsh feature that z-sy-h uses. Therefore, should a note about the OS-dependence of regexp syntax and a reference to the relevant OS-dependent manual page (re_format(7)?) be added to zsh's manual? The note could go in [the documentation of the `zsh/regex` module](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fregex-Module), for example. The zsh manual sources are in zsh's repository: https://github.com/zsh-users/zsh/tree/master/Doc/Zsh

Absolutely, I will try to move information to the Zsh manuals.

danielshahaf:
Thanks, LGTM!  I'll wait a bit for further feedback from others, and squash + merge as-is otherwise.

docs/highlighters/regexp.md
```diff
@@ -12,11 +12,32 @@ To use this highlighter, associate regular expressions with styles in the
...
+ZSH_HIGHLIGHT_REGEXP+=('^rm .*' fg="red",bold)
 ```
 
-This will highlight "sudo" only as a complete word, i.e., "sudo cmd", but not
-"sudoedit"
+This will highlight the whole line starting with `rm` command (for all
+operating systems, in contrast to the below example).
+
+Some regular expressions are [subject to the host platform][MAN_ZSH_REGEX], 
+especially the kernel. To highlight `sudo` only as a complete word, i.e.,
```
phy1729, line 22: Since the zsh/regex module uses `regexec`, the dependency is on the platform's libc not the kernel. (e.g. Debian/kFreeBSD which appears to have used glibc would support `\<` not `[[:<:]]`). I'd replace "especially the kernel" with "in particular the libc" or similar.

docs/highlighters/regexp.md
```diff
@@ -12,11 +12,32 @@ To use this highlighter, associate regular expressions with styles in the
...
-"sudoedit"
+This will highlight the whole line starting with `rm` command (for all
+operating systems, in contrast to the below example).
+
+Some regular expressions are [subject to the host platform][MAN_ZSH_REGEX], 
+especially the kernel. To highlight `sudo` only as a complete word, i.e., 
+`sudo cmd`, but not `sudoedit`, the respective regular expressions for the host
+systems would be:
+
+* GNU-Linux
```
phy1729, line 26: I believe this is usually stylized as `GNU/Linux`.

docs/highlighters/regexp.md
```diff
@@ -12,11 +12,32 @@ To use this highlighter, associate regular expressions with styles in the
 
 ```zsh
 typeset -A ZSH_HIGHLIGHT_REGEXP
-ZSH_HIGHLIGHT_REGEXP+=('\bsudo\b' fg=123,bold)
+ZSH_HIGHLIGHT_REGEXP+=('^rm .*' fg="red",bold)
```
phy1729, line 15: Quotes around red aren't needed and are a little odd, but just a style nit.

docs/highlighters/regexp.md
danielshahaf: Good point.  It _is_ usually stylized as you say, but let's be more accurate and state "GNU libc (e.g., most Linux distros)", as per your other comment.  The other bullet can remain as-is because on BSD platforms libc isn't as independent/swappable a component as it is on Linux.

docs/highlighters/regexp.md
jnooree: Thank you for correction. I'll change this part.

docs/highlighters/regexp.md
jnooree: > Good point. It _is_ usually stylized as you say, but let's be more accurate and state "GNU libc (e.g., most Linux distros)", as per your other comment. The other bullet can remain as-is because on BSD platforms libc isn't as independent/swappable a component as it is on Linux.

I will also take this part! Thank you both.

docs/highlighters/regexp.md
jnooree: Well, that's because I also work with python. Thank you very much!

danielshahaf:
@jnooree Thanks for the revisions.

@phy1729 Are we clear to merge, or?

The only issue I see is the phrase "many GNU/Linux".  "GNU/Linux" is an adjective or a singular noun and "many" should be followed by a plural noun, such as "distributions".

jnooree:
> The only issue I see is the phrase "many GNU/Linux". "GNU/Linux" is an adjective or a singular noun and "many" should be followed by a plural noun, such as "distributions".

What a shame! I changed it.

jnooree:
I have found something more about this. After adding this to ~/.zshrc:
```zsh
autoload zsh/pcre
setopt RE_MATCH_PCRE
```
then
```zsh
$ [[ "sudo" =~ "\bsudo\b" ]] && echo true || echo false
> true # before, it was false
```
I think it would be better adding the two lines to the docs rather than explaining everything,
as @danielshahaf mentioned earlier.

danielshahaf:
The `emulate` call will have reset that option to its default value by the time the regexp is compiled/matched, won't it?

Also, PCRE is optional at compile time.

Adding support for PCRE patterns sounds like a fair feature request, though.

danielshahaf:
My first question refers to what happens when the `regexp` highlighter compiles a string for matching.

jnooree:
> The `emulate` call will have reset that option to its default value by the time the regexp is compiled/matched, won't it?
> My first question refers to what happens when the `regexp` highlighter compiles a string for matching.

Actually, when I test the "word boundary" syntax, the highlighter has evaluated it correctly.
 
> Also, PCRE is optional at compile time.

But I didn't know about this. Then it would be better to leave the docs.

danielshahaf:
> Actually, when I test the "word boundary" syntax, the highlighter has evaluated it correctly.

I stand corrected.

> But I didn't know about this. Then it would be better to leave the docs.

Well, that's one option.  Another option is to mention in the docs that _if_ your build has the pcre module available, then you can use it to get PCRE features (and a platform-independent syntax to boot); I guess most distro builds do ship that module.  However, see #748 which I just opened for a potential can of worms here.

> I have found something more about this. After adding this to ~/.zshrc:
> 
> ```shell
> autoload zsh/pcre
> setopt RE_MATCH_PCRE
> ```

Did this test happen on the same machine you saw [the zmodload error](https://github.com/zsh-users/zsh-syntax-highlighting/issues/746#issuecomment-652373045) on?  Asking this in context of whether most distro packages ship zsh with PCRE support.

jnooree:
> Did this test happen on the same machine you saw [the zmodload error](https://github.com/zsh-users/zsh-syntax-highlighting/issues/746#issuecomment-652373045) on? Asking this in context of whether most distro packages ship zsh with PCRE support.

Yes, it did. All I need was the first line, ```autoload zsh/pcre```. I think the error comes out if the module is not loaded, but not if the module doesn't exist.

danielshahaf:
On Linux, I don't get dlopen() errors when the module is available but not loaded:

```sh
% zsh -fc '[[ foo -pcre-match bar ]]' 
zsh:1: unknown condition: -pcre-match
zsh: exit 2     zsh -fc '[[ foo -pcre-match bar ]]'
% zsh -fc 'zmodload zsh/pcre; [[ foo -pcre-match bar ]]' 
zsh: exit 1     zsh -fc 'zmodload zsh/pcre; [[ foo -pcre-match bar ]]'
```

Nor in a static build where pcre was disabled at build time:

```sh
% [[ foo -pcre-match bar ]] 
zsh: unknown condition: -pcre-match
```

So I'm not sure what to make of the error you saw.

By the way, you keep writing `autoload`.  I assume you mean `zmodload`.  `autoload zsh/pcre` simply marks the function called `zsh/pcre` as autoloadable; it doesn't have anything to do with the module.

jnooree:
> On Linux, I don't get dlopen() errors when the module is available but not loaded:
> 
> ```shell
> % zsh -fc '[[ foo -pcre-match bar ]]' 
> zsh:1: unknown condition: -pcre-match
> zsh: exit 2     zsh -fc '[[ foo -pcre-match bar ]]'
> % zsh -fc 'zmodload zsh/pcre; [[ foo -pcre-match bar ]]' 
> zsh: exit 1     zsh -fc 'zmodload zsh/pcre; [[ foo -pcre-match bar ]]'
> ```
> 
> Nor in a static build where pcre was disabled at build time:
> 
> ```shell
> % [[ foo -pcre-match bar ]] 
> zsh: unknown condition: -pcre-match
> ```
> 
> So I'm not sure what to make of the error you saw.
>
> By the way, you keep writing `autoload`. I assume you mean `zmodload`. `autoload zsh/pcre` simply marks the function called `zsh/pcre` as autoloadable; it doesn't have anything to do with the module.

Now I've got the point. I changed the shell from the macOS system-wide ```zsh``` (```/bin/zsh```) from the ```homebrew``` version of ```zsh``` (during this conversation). Same error still exists with the system's ```zsh```, but no problem with the ```homebrew``` version. Apple might have done something to the shell.

danielshahaf:
**Summary:** The patch is correct for the case that the RE_MATCH_PCRE option is unset, but doesn't cover the case that that option _is_ set.

We're going to have to document the dependency on that option anyway, regardless of how we resolve #748.

So, what do you think of the following?  It's @jnooree's text, extended to cover PCRE and copyedited.

````markdown
This will highlight lines that start with a call to the `rm` command.

The regular expressions flavour used is [PCRE][pcresyntax] when the
`RE_MATCH_PCRE` option is set and POSIX Extended Regular Expressions (ERE),
as implemented by the platform's C library, otherwise.  For details on the
latter, see [the `zsh/regex` module's documentation][MAN_ZSH_REGEX] and the
`regcomp(3)` and `re_format(7)` manual pages on your system.

For instance, to highlight `sudo` only as a complete word, i.e., `sudo cmd`,
but not `sudoedit`, one might use:

* When the `RE_MATCH_PCRE` is set:

    ```zsh
    typeset -A ZSH_HIGHLIGHT_REGEXP
    ZSH_HIGHLIGHT_REGEXP+=('\bsudo\b' fg=123,bold)
    ```

* When the `RE_MATCH_PCRE` is unset, on platforms with GNU `libc` (e.g., many GNU/Linux distributions):

    ```zsh
    typeset -A ZSH_HIGHLIGHT_REGEXP
    ZSH_HIGHLIGHT_REGEXP+=('\<sudo\>' fg=123,bold)
    ```

* When the `RE_MATCH_PCRE` is unset, on BSD-based platforms (e.g., macOS):

    ```zsh
    typeset -A ZSH_HIGHLIGHT_REGEXP
    ZSH_HIGHLIGHT_REGEXP+=('[[:<:]]sudo[[:>:]]' fg=123,bold)
    ```

Note, however, that PCRE and POSIX ERE have a large common subset:
for instance, the regular expressions `[abc]`, `a*`, and `(a|b)` have the same
meaning in both flavours.

[pcresyntax]: http://www.pcre.org/original/doc/html/pcresyntax.html
````

danielshahaf:
Merged with @phy1729's review on https://github.com/zsh-users/zsh-syntax-highlighting/issues/722#issuecomment-1047427725.  Thank you very much, @jnooree, and sorry we hadn't gotten to this sooner!

PR #751, opened 2020-07-19
Bug fix: infinite loop when the current working directory doesn't exist

romkatv:
To reproduce:

```zsh
mkdir /tmp/foo
cd /tmp/foo
rmdir /tmp/foo
exec zsh -f
source ~/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
```

Press <kbd>a</kbd> and zle will hang.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -1211,14 +1211,18 @@ _zsh_highlight_main_highlighter_check_path()
   if [[ $expanded_path[1] == / ]]; then
     tmp_path=$expanded_path
   else
-    tmp_path=$PWD/$expanded_path
+    # Unlike $PWD, ${(%):-%/} predictably expands to '.' if the current
+    # working directory doesn't exist.
+    tmp_path=${(%):-%/}/$expanded_path
   fi
   tmp_path=$tmp_path:a
 
-  while [[ $tmp_path != / ]]; do
-    [[ -n ${(M)ZSH_HIGHLIGHT_DIRS_BLACKLIST:#$tmp_path} ]] && return 1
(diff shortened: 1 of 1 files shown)


danielshahaf:
Thanks.  Triaged.  Milestoned 0.9.0 because it's not a regression, though realistically I'd like to see it in 0.8.0 or, failing that, 0.8.1.  No time to test or review right now, sorry.

phy1729:
@romkatv thank you for the minimal reproducer. Thoughts on
```diff
diff --git a/highlighters/main/main-highlighter.zsh b/highlighters/main/main-highlighter.zsh
index e59c61c..24d0b75 100644
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -1210,6 +1210,9 @@ _zsh_highlight_main_highlighter_check_path()
   # Check if this is a blacklisted path
   if [[ $expanded_path[1] == / ]]; then
     tmp_path=$expanded_path
+  elif ! [[ -e . ]]; then
+    # If $PWD doesn't exist, relative paths are impossible
+    return 1
   else
     tmp_path=$PWD/$expanded_path
   fi
```

Am I missing an edge case?

romkatv:
> No time to test or review right now, sorry.

This is not blocking me in any way (I never delete my current directory), so take your time.

> @romkatv thank you for the minimal reproducer. Thoughts on

Your version of the patch has a race condition (so it can still go into an infinite loop) and doesn't colorize any files if the current directory doesn't exist (even though some files could still be colorized). That said, I've very little emotional investment in this PR one way or another, so if you decide to apply your patch I won't argue.

PR #757, opened 2020-08-09
Fix test suite on Cygwin

m0vie:
'make test' fails under Cygwin:

```
$ make test
ZSH_PATCHLEVEL=zsh-5.8-243-g5b4cfb1
Running test brackets
env: ‘zsh’: No such file or directory
Running test main
env: ‘zsh’: No such file or directory
Running test pattern
env: ‘zsh’: No such file or directory
Running test regexp
env: ‘zsh’: No such file or directory
make: *** [Makefile:40: test] Error 127
```

This is because on Cygwin, env -i zsh does not work as expected (Cygwin does not lookup the command in /usr/bin/ or /usr).

After fixing that, there are two more problems:
```
## parameter-to-global-alias
# BUFFER=$'$s'
not ok 1 - [1,2] «$s» - expected (1 2 "unknown-token"), observed (1 2 "command").
ok 2 - cardinality check
# expected_region_highlight  region_highlight
# '1 2 unknown-token'        '0 2 command'
```

```
## precommand-killing1
# BUFFER=$'sudo -e /etc/passwd'
ok 1 - [1,4] «sudo»
ok 2 - [6,7] «-e»
not ok 3 - [9,19] «/etc/passwd» - expected (9 19 "path"), observed (9 19 "default").
ok 4 - cardinality check
# expected_region_highlight   region_highlight
# '1 4 precommand'            '0 4 precommand'
# '6 7 single-hyphen-option'  '5 7 single-hyphen-option'
# '9 19 path'                 '8 19 default'
```


danielshahaf:
Thanks. Haven't reviewed yet, but question: Is there a Cygwin CI service we could use to catch such problems sooner?

highlighters/main/test-data/parameter-to-global-alias.zsh
```diff
@@ -28,8 +28,8 @@
 # vim: ft=zsh sw=2 ts=2 et
 # -------------------------------------------------------------------------------------------------
 
-alias -g x=y
-local s=x
+alias -g global_alias_7cc33aff-18d4-48cd-8f97-38c272119a53=y
+local s=global_alias_7cc33aff-18d4-48cd-8f97-38c272119a53
```
danielshahaf, line 32: Could you please either document the magic number, or use a self-documenting name, or have the test skip when its assumptions don't hold (see `$skip_test`)?

highlighters/main/test-data/precommand-killing1.zsh
```diff
@@ -30,10 +30,10 @@
 
 hash sudo=false
 
-BUFFER='sudo -e /etc/passwd'
+BUFFER='sudo -e /usr'
```
danielshahaf, line 33: On this value of $BUFFER, since $CURSOR isn't equal to $#BUFFER, it would be correct to flag `/usr` as `unknown-token`, even if the code doesn't do this yet.

I suppose we could `touch foo` and then use `BUFFER='sudo -e ./foo'`, couldn't we?

Makefile
```diff
@@ -41,7 +41,7 @@ test:
 	for test in highlighters/*; do \
 		if [ -d $$test/test-data ]; then \
 			echo "Running test $${test##*/}"; \
-			env -i QUIET=$$QUIET $${TERM:+"TERM=$$TERM"} $(ZSH) -f tests/test-highlighting.zsh "$${test##*/}"; \
+			env -i QUIET=$$QUIET $${TERM:+"TERM=$$TERM"} PATH="$$PATH" $(ZSH) -f tests/test-highlighting.zsh "$${test##*/}"; \
```
danielshahaf, line 44: This would leave $PATH set for the entirety of the test run, though.  How about:

```diff
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ INSTALL?=install -c
 PREFIX?=/usr/local
 SHARE_DIR?=$(DESTDIR)$(PREFIX)/share/$(NAME)
 DOC_DIR?=$(DESTDIR)$(PREFIX)/share/doc/$(NAME)
-ZSH?=zsh # zsh binary to run tests with
+ZSH?="`which zsh`" # zsh binary to run tests with
 
 all:
        cd docs && \
```

?

highlighters/main/test-data/parameter-to-global-alias.zsh
m0vie: I thought using a GUID was self-expanatory, but I guess we can just use something like
`global_alias_which_hopefully_does_not_clash_with_an_external_command_name`

danielshahaf: The problem with a UUID is that it's not clear whether it was a fresh nonce or a well-known value.

highlighters/main/test-data/precommand-killing1.zsh
m0vie: `touch foo` sounds like the more natural solution, yep.

Makefile
m0vie: Doens't work:
```
$ make test
ZSH_PATCHLEVEL=zsh-5.8-243-g5b4cfb1
Running test brackets
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
Running test main
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
Running test pattern
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
Running test regexp
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
/usr/bin/env: 'zsh': No such file or directory
make: *** [Makefile:40: test] Error 2
```
Reason: The shebang of `test-highlighting.zsh` contains `#!/usr/bin/env zsh`, i.e. another `env` call, which then again won't find `zsh` anymore if `PATH` is not set.

I guess we are better of with having `PATH` set. Potential external commands inside test-cases itself also need to be found.

m0vie:
> Is there a Cygwin CI service we could use to catch such problems sooner?

I am not aware of any.

Makefile
danielshahaf: The shebang line? Really? I'd have expected it not to matter when `-c` isn't given.

How about `$(ZSH) -fc '. tests/test-highlighting.zsh` to bypass the shebang line? Would that break anything?

highlighters/main/test-data/precommand-killing1.zsh
danielshahaf: LGTM

Makefile
danielshahaf: > The shebang line? Really? I'd have expected it not to matter when `-c` isn't given.

On Linux on FreeBSD it doesn't matter.  What's the output of `zsh -f =(<<<$'#!/usr/bin/env date')` on Cygwin?

highlighters/main/test-data/precommand-killing1.zsh
```diff
@@ -29,11 +29,12 @@
...
+touch foo
 
-BUFFER='sudo -e /etc/passwd'
+BUFFER='sudo -e ./foo'
 
 expected_region_highlight=(
   '1 4 precommand' # sudo
   '6 7 single-hyphen-option' # -e
-  '9 19 path' # /etc/passwd
+  '9 13 path' # ./foo
```
danielshahaf, line 39: Cherry-picked to master: 79b6e7e05f96.

highlighters/main/test-data/parameter-to-global-alias.zsh
```diff
@@ -28,8 +28,8 @@
 # vim: ft=zsh sw=2 ts=2 et
 # -------------------------------------------------------------------------------------------------
 
-alias -g x=y
-local s=x
+alias -g global_alias_which_hopefully_does_not_clash_with_an_external_command_name=y
+local s=global_alias_which_hopefully_does_not_clash_with_an_external_command_name
```
danielshahaf, line 32: Deployed a partial fix in 6d5372a937c63a565a1ff93aa6efa9a2b6179917.  If you'd like to improve things further, you're very welcome to do so, but do so in a new issue or pull request — do not overload this one.

m0vie: I don't like the solution of 6d5372a. This just skips the test now for no good reason.

Avoiding the name of the alias to be a command could at least be avoided by also choosing something other than `x` which is not as likely to clash.

Makefile
m0vie: Correction: It is not the shebang of `test-highlighting.zsh` but the other external scripts at are being run *from* there. E.g.
```
print > >($results_filter | $root/tests/tap-colorizer.zsh) -r -- "# global (driver) tests"
print > >($results_filter | $root/tests/tap-colorizer.zsh) -r -- "1..1"
```

What this actually means is that if ZSH is set in the Makefile, this does not guarantee that every part of the test suite is executed with that exact ZSH binary. Those called scripts might not be.

danielshahaf: tap-colorizer needn't be run under the same zsh version as everything else. The only reason it's written in zsh is to minimize dependencies.

Anyway, feel free to PR the fix.

highlighters/main/test-data/parameter-to-global-alias.zsh
m0vie: #761

danielshahaf:
@m0vie Ping?

m0vie:
I have already pushed an updated commit (improved commit message) to this PR.

What else is needed?

danielshahaf:
Address https://github.com/zsh-users/zsh-syntax-highlighting/pull/757/files#r467688965.

m0vie:
There is nothing really to address.

Keeping `PATH` completely unset is not an option on Cygwin, because the default fallback of `/bin` and `/usr/bin` won't be searched, like e.g. on Linux.

That means that every (external) utility that is being invoked with a shebang in the test code will fail.

The only thing I can think of is something along the line of

```
env -i ... PATH=/usr/bin:/bin ...
```

i.e, to force the fallback behavior of most platforms explicitly.

Then again, I see no harm in simply passing PATH along unchanged. You'd have to have a really weird development environment for that to break anything, wouldn't you? Travis seems to agree.

danielshahaf:
> Keeping `PATH` completely unset is not an option on Cygwin, because the 
> default fallback of `/bin` and `/usr/bin` won't be searched, like e.g. 
> on Linux.
> 

Are you sure about this?  What about, say, `env -i /bin/bash` and therein (interactively) running `env`?  What does that print?

> That means that every (external) utility that is being invoked with a 
> shebang in the test code will fail.
> 

Why are you focussing on shebangs?  Those are trivial to solve: just prepend «$ZSH» to the call (or skip the call entirely — tap-filter and tap-colorizer should both be optional).  The real problem is external programs such as cat(1), rm(1), sed(1).  Are those being found (on current master)?  If so, how?  If not, how did the test suite ever work on Cygwin?

> The only thing I can think of is something along the line of
> 
> `env -i ... PATH=/usr/bin:/bin ...
> ` 
> i.e, to force the fallback behavior of most platforms explicitly.
> 

It's not our place to hardcode every platform's default $PATH, but perhaps we could derive it?  Something along the lines of «PATH=`ZDOTDIR=$(mktemp -d) PATH= $(which zsh) -c 'print -r -- ${(q-)PATH}'`» might do the trick.

> Then again, I see no harm in simply passing PATH along unchanged. You'd 
> have to have a really weird development environment for that to break 
> anything, wouldn't you? Travis seems to agree.

The problem statement given was "'env -i zsh' doesn't find zsh".  To fix that by passing $PATH through to all tests is to kill a fly with a cannon.

m0vie:
> Are you sure about this? What about, say, `env -i /bin/bash` and therein (interactively) running `env`? What does that print?

```
zsh:~$ env -i /bin/bash
bash:~$ env
LS_COLORS=
BLOCK_SIZE='1
DISPLAY=:0.0
SYSTEMROOT=C:\WINDOWS
PWD=/home/mv
DF_BLOCK_SIZE='M
WINDIR=C:\WINDOWS
SHLVL=1
EXECIGNORE=*.dll
_=/usr/bin/env
bash:~$ env bash
env: 'bash': No such file or directory
```

> The real problem is external programs such as cat(1), rm(1), sed(1). Are those being found (on current master)?  If so, how? 

They are. The problem is only with 'env' itself behaving very strangely (hence the focus on the shebang of the scripts containing it). Sorry, if that was not clear from the beginning, I didn't fully understand it myself up to now.

> Those are trivial to solve: just prepend «$ZSH» to the call.

You are right, that does work. (Just need to pass `ZSH` via `env -i ... ZSH=${ZSH} ...` in the `Makefile`).

I have pushed the branch again. Tests rune fine on Cygwin using that approach. What do you think?

danielshahaf:
> They are. The problem is only with 'env' itself behaving very strangely (hence the focus on the shebang of the scripts containing it). Sorry, if that was not clear from the beginning, I didn't fully understand it myself up to now. 

Wait, so you're saying that within `env -i zsh`, $PATH is unset, and plain `sed` works but literally `env sed` fails?  Why does the former work?  What's `type -w sed` at that point?  Might the difference be a Cygwin bug?

m0vie:
> Wait, so you're saying that within `env -i zsh`, $PATH is unset, and plain `sed` works but literally `env sed` fails?

Yes, it seems to be due to the fact that
1) `env -i` removes PATH from the environment
2) When invoking, in `/etc/zshenv` PATH at least gets set to some sane default again
3) But that sane default is *not* exported, so while zsh is able to directly invoke commands again, when invoking `env` again from the shell it is again fully unset and no commands are found

So this has to do with the default lookup fallback after all.

One could argue that this is a bug in Cygwin, indeed. However, this appears to be know for over 17 years:
-https://cygwin.cygwin.narkive.com/aMNM4qT7/env-xargs-no-such-file-or-directory 
-https://sourceware.org/legacy-ml/cygwin/2004-09/msg00128.html
-http://mailman.uclinux.org/pipermail/uclinux-dev/2005-May/032483.html

We can work around that by making sure `PATH` is exported from `tests/test-highlighting.zsh` for command invocations from there. See updated commit.

danielshahaf:
Point (1) isn't a bug; it's a POSIX-compliant behaviour.

Point (3) sounds like it might be a bug, and doesn't seem to be covered by the references you listed. Why don't you raise it with the Cygwin zsh package maintainers?

Could you make the `export` conditional on running on Cygwin.

Issue #536, opened 2018-09-22
Canonicalize option settings earlier


danielshahaf:
#535 / 3a7ddcd was a bug due to `_zsh_highlight_load_highlighters` running with the interactive environment's options. We already call `emulate -L` in `_zsh_highlight` for precisely this reason.

Long-term we should switch to an autoload design (#252), but in the interim, we should review the driver to see if we can either:

- Run more of the driver's code under canonicalized option settings; or
- Review the driver's code to see if it's robust for being run under _arbitrary_ setopt's (particularly including those that change syntax)

danielshahaf:
Cross-referencing #688.

danielshahaf:
Triage: not a blocker; depends on #252/#688 which depend on redrawhook which depends on https://github.com/zsh-users/zsh/pull/57; so: deferring to 0.9.0 and labeling so that once redrawhook is merged we'll re-triage this.

danielshahaf:
Re-triaged.  The dependencies #252/#688 are milestone 0.9.0, therefore so is this issue.  Removing the temporary label (see previous comment).

Issue #252, opened 2015-12-20
Consider converting to an `autoload`-able function


danielshahaf:
Consider converting zsh-syntax-highlighting from a `source`d script to an `autoload`ed function.  Benefits include better isolation from calling environment and `zcompile` byte-compliation support.

danielshahaf:
Using `autoload -U` would prevent aliases in the calling environment from affecting z-sy-h.  (See for example the following downstream bug report: <https://bugs.debian.org/842994>.)

danielshahaf:
Note also the [workers/40327](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=40327) changes to `autoload` with an absolute path argument.

danielshahaf:
[workers/40458](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=40458) mentions a useful "autoload a number of related functions" that's now possible in 5.4-to-be.

phy1729:
Will the highlighter interface change to allow for autoloaded highlighters?

danielshahaf:
From a mile high view, I assume that'd be an improvement.

danielshahaf:
Once that's done we can also `emulate -L` and `setopt warncreateglobal ${(k)zsyh_user_options[(I)warnnestedglobal}` as in #413.  (That construct expands to `"warnnestedglobal"` on zsh's that have that option and to nothing otherwise.)

danielshahaf:
Cross-referencing #688.

danielshahaf:
Triage: Trivial to implement (see https://github.com/okapia/zsh-viexchange/blob/339dc2d8c4d8a164c7b441d0c1e6442b0b38b3d0/zsh-viexchange.plugin.zsh#L1-L4), will help with #688 (resilience to user's setopts) and #635/#734 (loading performance): milestone 0.8.0 optimistically, but may defer this to 0.9.0 after all (e.g., if this turns out to have to be merged after redrawhook).

danielshahaf:
> milestone 0.8.0 optimistically, but may defer this to 0.9.0 after all (e.g., if this turns out to have to be merged after redrawhook).

It _would_ be useful to merge this only after redrawhook, simply because that's the easiest way to avoid spurious conflicts in that merge.  (Fixing this will involve renaming the entry point file and creating a stub in its stead; git would then conflict, rather than merge conflict-freely to the renamed file.)

Issue #756, opened 2020-08-09
conflict with zsh-autosuggestions


leon728:
when loading both (zsh-autosuggestions and zsh-syntax-highlighting), text color of zsh-autosuggestions is not darkgray.
when loading only zsh-syntax-highlighting, evething works fine.
(reference: https://github.com/zsh-users/zsh-autosuggestions/issues/516)

it was caused by latest commit, using 2d60a47 works fine.

danielshahaf:
How could https://github.com/zsh-users/zsh-autosuggestions/issues/516 be related to this? It's from March and the latest commit is from yesterday. Furthermore, that issue doesn't mention what branch or tag of zsh-syntax-highlighting was used. I assume you just mean your symptoms are the same as the screenshots in that issue.

The latest commit is the redrawhook merge. This issue was reported before the merge and believed fixed then: see issues #579 and #735. I'll look into this again.

Please report the version numbers of zsh and zsh-autosuggestions.

leon728:
Yes, I mean symptoms are the same as https://github.com/zsh-users/zsh-autosuggestions/issues/516

ZSH version:
zsh 5.8 (x86_64-ubuntu-linux-gnu)

zsh-autosuggestions version:
https://github.com/zsh-users/zsh-autosuggestions/commit/ae315ded4dba10685dbbafbfa2ff3c1aefeb490d

zsh-syntax-highlighting version:
https://github.com/zsh-users/zsh-syntax-highlighting/commit/2d60a47cc407117815a1d7b331ef226aa400a344 : ok
https://github.com/zsh-users/zsh-syntax-highlighting/commit/c14fcad3b08e19dce750b4b0eb6cc597e0f93d57 : failed

you fixed https://github.com/zsh-users/zsh-syntax-highlighting/issues/579 in https://github.com/zsh-users/zsh-syntax-highlighting/commit/cb33cc0081234c0f8e48d89e0ba81cf58a8f7635
but in my test, https://github.com/zsh-users/zsh-syntax-highlighting/commit/9ce3540d74c12b24a3f705a115f1f7b8f997ced6 is ok, https://github.com/zsh-users/zsh-syntax-highlighting/commit/cb33cc0081234c0f8e48d89e0ba81cf58a8f7635 failed

reproduce (https://github.com/zsh-users/zsh-syntax-highlighting/issues/579):
`exec zsh -f`
`source path/to/zsh-autosuggestions/zsh-autosuggestions.zsh`
[...]

PR #758, opened 2020-08-10
Autoload and `emulate zsh`

phy1729:
Since redrawhook is now merged, figured I'd start the autoload highlighters and emulate zsh source driver branch.

Still appears to work locally (including `setopt interactivecomments` and then typing a line with comments). I'm running 5.8 though, so tests with a dev version with redrawhook active are needed.

Related issues: #252 #536 #688


phy1729:
Tests pass on `3423a97 driver: Split out autoloadable part of _zsh_highlight`, but fail on `ef754de driver: Run under emulate zsh`. I believe this is due to defining a function under emulate being sticky, but will need to further investigate later.

danielshahaf:
> Tests pass on `3423a97 driver: Split out autoloadable part of _zsh_highlight`, but fail on `ef754de driver: Run under emulate zsh`. I believe this is due to defining a function under emulate being sticky, but will need to further investigate later.
> 

Noted.  Let me know when you'd like me to run tests under master.

danielshahaf:
I mean, under zsh master. :)

highlighters/main/_zsh_highlight_highlighter_main_paint
```diff
@@ -0,0 +1,1756 @@
...
+{
+  (( $# == 1 )) || print -r -- >&2 "zsh-syntax-highlighting: BUG: _zsh_highlight_main_highlighter_expand_path: called without argument"
+
+  # The $~1 syntax normally performs filename generation, but not when it's on the right-hand side of ${x:=y}.
+  setopt localoptions nonomatch
+  unset REPLY
+  : ${REPLY:=${(Q)${~1}}}
+}
+
+_zsh_highlight_highlighter_main_paint
```
danielshahaf, line 1756: Add `"$@"`?

highlighters/cursor/_zsh_highlight_highlighter_cursor_predicate
```diff
@@ -0,0 +1,32 @@
+# -------------------------------------------------------------------------------------------------
+# Copyright (c) 2010-2011 zsh-syntax-highlighting contributors
```
danielshahaf, line 2: Kudos on setting these correctly!

danielshahaf:
Skimmed.  It's very comprehensive.  Much kudos!  Let me know how I can help.

It's an invasive change so I'll definitely prefer to merge it _before_ a release rather than after, to ease backports, should any be needed.  I take it you're targeting this branch for 0.8.0.  (I don't have a strong opinion one way or the other.)

zsh-syntax-highlighting.zsh
```diff
@@ -27,571 +27,9 @@
...
-if (( $+X_ZSH_HIGHLIGHT_DIRS_BLACKLIST )); then
-  print >&2 'zsh-syntax-highlighting: X_ZSH_HIGHLIGHT_DIRS_BLACKLIST is deprecated. Please use ZSH_HIGHLIGHT_DIRS_BLACKLIST.'
-  ZSH_HIGHLIGHT_DIRS_BLACKLIST=($X_ZSH_HIGHLIGHT_DIRS_BLACKLIST)
-  unset X_ZSH_HIGHLIGHT_DIRS_BLACKLIST
-fi
-
-# Restore the aliases we unned
-eval "$zsh_highlight__aliases"
-builtin unset zsh_highlight__aliases
+emulate zsh -c 'source ${0:h}/driver.zsh'
```
danielshahaf, line 32: An interim commit used `0=${(%):-%N}; fpath+=( ${0:a:h} )`.  For the record, I'm not sure that interim state was correct in all cases, because `:a`, unlike `:P`, is not guaranteed to return something that refers to the same inode (or even to _an_ inode).  I think the compatible way to do this would have been to use ${0:P:h} if available, and `$(cd -q -- ${0:h} && pwd -P)` otherwise.

But that was just an interim state, and I'm mentioning this just for completeness.  I'm not suggesting to rewrite the history over this.

zsh-syntax-highlighting.zsh
```diff
@@ -27,571 +27,9 @@
...
-  # This disables highlighting during isearch (for reasons explained in README.md) unless zsh is new enough
-  # and doesn't have the pattern matching bug
-  if [[ $WIDGET == zle-isearch-update ]] && { $zsh_highlight__pat_static_bug || ! (( $+ISEARCHMATCH_ACTIVE )) }; then
-    return $ret
-  fi
-
-  # Before we 'emulate -L', save the user's options
-  local -A zsyh_user_options
-  if zmodload -e zsh/parameter; then
-    zsyh_user_options=("${(kv)options[@]}")
```
danielshahaf, line 195: This code now happens after the `emulate zsh -c 'source …/driver.zsh'`.  Is that a problem?

zsh-syntax-highlighting.zsh
```diff
@@ -27,571 +27,9 @@
...
-        # In either case, fall back to a version check.
-        #
-        # The memo= feature was added to zsh in commit zsh-5.8-172-gdd6e702ee.
-        # The version number at the time was 5.8.0.2-dev (see Config/version.mk).
-        # Therefore, on 5.8.0.3 and newer the memo= feature is available.
-        #
-        # On zsh version 5.8.0.2 between the aforementioned commit and the
-        # first Config/version.mk bump after it (which, at the time of writing,
-        # is yet to come), this condition will false negative.
-        if is-at-least 5.8.0.3; then
```
danielshahaf, line 166: I just pushed 3944a44ffe44 which changes two lines, including this one.  Sorry to have caused you conflicts, but it was a hotfix needed to unbreak z-asug under 5.8; see #756.

zsh-syntax-highlighting.zsh
phy1729: Indeed it is. Even though _zsh_highlight is not called under `emulate zsh`; definition under `emulate zsh -c` is sticky. I've moved it back into zsh-syntax-highlighting.zsh.

highlighters/main/_zsh_highlight_highlighter_main_paint
phy1729: Debated adding it, but it wasn't needed and I had already written the line. Doesn't hurt though and could avoid future bugs I suppose. Added.

zsh-syntax-highlighting.zsh
phy1729: No worries. This will create conflicts for everything else and bug fixes come before features. Rebased.

zsh-syntax-highlighting.zsh
phy1729: `fpath+=(${0:a:h})` is still used just in driver.zsh now. Some resolution needs to happen to avoid a relative path (which causes the tests to fail).

Do any of the pathological `:a` cases apply here with `$0`?

danielshahaf: I think so; watch:

```sh
% file /tmp/{slash,foo,foo/bar}
/tmp/slash:   symbolic link to /
/tmp/foo:     directory
/tmp/foo/bar: Paul Falstad's zsh script, ASCII text executable
% cat /tmp/slash/../tmp/foo/bar
#!/bin/zsh -f
echo "[$0]" I am /tmp/foo/bar
% source /tmp/slash/../tmp/foo/bar 
[/tmp/slash/../tmp/foo/bar] I am /tmp/foo/bar
% echo ${${:-/tmp/slash/../tmp/foo/bar}:a}
/tmp/tmp/foo/bar
% 
```

phy1729: Fixed to use `:P`.

phy1729:
Upon further testing, while tests may pass, interactively options that are reset by `emulate zsh` (e.g. equals) are reset by the callers of _zsh_highlight (in particular _zsh_highlight_call_widget pre-redrawhook. I've moved the required code back into zsh-syntax-highlighting.zsh so that _zsh_highlight still executes under the user's environment.

Still need to add the fallback code if `:P` isn't available, but I think that's it, so it's testable now.

danielshahaf:
Travis is failing on pre-5.3 zsh with several different warnings.  I didn't check whether some of these happen on 5.3-and-later too, which are green.

```tap
Running test pattern
# global (driver) tests
1..1
./driver.zsh:32: unrecognized modifier `P'
zsh:122: command not found: is-at-least
ok 1 # 'alias -- +foo=bar' is preserved
(anon):55: array parameter ZSH_HIGHLIGHT_HIGHLIGHTERS created globally in function (anon)
(anon):243: array parameter ZSH_HIGHLIGHT_STYLES created globally in function (anon)
1..2
## rm-rf
# BUFFER=$'rm -rf /'
not ok 1 - unmatched expectation (1 8 fg=white,bold,bg=red) 
todo=''
not ok 2 - cardinality check - have 1 expectations and 0 region_highlight entries: Â«expected_region_highlight=( $'1 8 fg=white,bold,bg=red' )Â» Â«region_highlight=(  )Â»
# expected_region_highlight   region_highlight
# '1 8 fg=white,bold,bg=red'
/work/./highlighters/pattern/test-data/rm-rf.zsh:30: array parameter ZSH_HIGHLIGHT_PATTERNS created globally in function (anon)
_zsh_highlight:51: command not found: _zsh_highlight_internal
Bail out! On './highlighters/pattern/test-data/rm-rf.zsh': output on stderr
Running test regexp
# global (driver) tests
1..1
./driver.zsh:32: unrecognized modifier `P'
zsh:122: command not found: is-at-least
ok 1 # 'alias -- +foo=bar' is preserved
(anon):55: array parameter ZSH_HIGHLIGHT_HIGHLIGHTERS created globally in function (anon)
(anon):243: array parameter ZSH_HIGHLIGHT_STYLES created globally in function (anon)
1..3
## complex
# BUFFER=$'echo 1+9-3 7+2'
not ok 1 - unmatched expectation (6 10 fg=white,bold,bg=red) 
not ok 2 - unmatched expectation (12 14 fg=white,bold,bg=red) 
todo=''
not ok 3 - cardinality check - have 2 expectations and 0 region_highlight entries: Â«expected_region_highlight=( $'6 10 fg=white,bold,bg=red' $'12 14 fg=white,bold,bg=red' )Â» Â«region_highlight=(  )Â»
# expected_region_highlight     region_highlight
# '6 10 fg=white,bold,bg=red'
# '12 14 fg=white,bold,bg=red'
/work/./highlighters/regexp/test-data/complex.zsh:30: array parameter ZSH_HIGHLIGHT_REGEXP created globally in function (anon)
_zsh_highlight:51: command not found: _zsh_highlight_internal
Bail out! On './highlighters/regexp/test-data/complex.zsh': output on stderr
Makefile:39: recipe for target 'test' failed
make: *** [test] Error 2
travis_time:end:03040b10:start=1597206903543376108,finish=1597206911136524385,duration=7593148277,event=script

The command "docker run -v $PWD:/work -w /work zshusers/zsh:${ZSH} /bin/sh -c 'install_packages make procps bsdmainutils && make test'" exited with 2.
```

danielshahaf:
Tested under zsh zsh-5.8-177-gd14a924 and zsh-5.8-241-g7facca6.  All tests pass under both.  Starting the latter interactively, sourcing z-sy-h, unsetting EQUALS, then typing `echo foo =ls bar`, results in `=ls` being underlined, though it shouldn't be.  What else should I test?

issue-dispenser:
@danielshahaf @phy1729 What are the remaining blockers for this to be merged?

PR #760, opened 2020-08-10
tests: Adapt for off-by-one difference in start offset in expectations

### LINKED ISSUE #759: test: start offset in expected_region_highlight is off-by-one to the actual format of zsh's region_highlight
The actual format of `region_highlight` is different to the defined expectations `expected_region_highlight` in tests:

E.g. for a region describing the full buffer `BUFFER=abcd` zsh would use `0 4 default`, but the expectations in test cases need to define `1 4 default`.

This is extremely confusing:

```
## cursor-matchingbracket
# BUFFER=$': ((( )))'
ok 1 - [3,3] «(»
ok 2 - [4,4] «(»
ok 3 - [5,5] «(»
ok 4 - [7,7] «)»
not ok 5 - [8,8] «)» - expected (8 8 "default"), observed (8 8 "bracket-level-2").
ok 6 - [9,9] «)»
ok 7 - [9,9] «)»
ok 8 - cardinality check
# expected_region_highlight     region_highlight
# '3 3 bracket-level-1'         '2 3 bracket-level-1'
# '4 4 bracket-level-2'         '3 4 bracket-level-2'
# '5 5 bracket-level-3'         '4 5 bracket-level-3'
# '7 7 bracket-level-3'         '6 7 bracket-level-3'
# '8 8 default'                 '7 8 bracket-level-2'
# '9 9 bracket-level-1'         '8 9 bracket-level-1'
# '9 9 cursor-matchingbracket'  '8 9 cursor-matchingbracket'
```

We should either 
a) Convert everything to the test-specific format before test assertions
-> Advantage: the test format is more human-readable
-> Disadvantage: this is not the format that zsh actually uses
b) decrement all the start offsets in every test case by 1.

m0vie:
Fixes #759.

### DIFF
--- a/tests/test-highlighting.zsh
+++ b/tests/test-highlighting.zsh
@@ -159,11 +159,19 @@ run_test_internal() {
   unset ARG
 
   integer print_expected_and_actual=0
+  integer i
 
   if (( unsorted )); then
     region_highlight=("${(@n)region_highlight}")
     expected_region_highlight=("${(@n)expected_region_highlight}")
   fi
+
+  # start offset in expected_region_highlight in test cases is off-by-one to the actual zsh-internal format of region_highlight
+  # convert region_highlight to test-specific format for better readability:
+  for ((i=1; i<=$#region_highlight; i++)); do
+    local -a highlight_zone; highlight_zone=( ${(z)region_highlight[i]} )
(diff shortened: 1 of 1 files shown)


danielshahaf:
**Triage:** Not ready to merge, because:

- CI is red.

- Documentation has not been updated.

- Concerns with the approach are being discussed on #759.

PR #761, opened 2020-08-11
tests: parameter-to-global-alias: Use alias name less likely to clash

m0vie:


danielshahaf:
Milestoned.

Would you link to the man page of the `x` command (lowercase)?

m0vie:
There is no man-page. It's just some script that comes with the cygwin-port of `xorg-server`.

```
$ man x
No manual entry for x

$ cygcheck -f =x
xorg-server-1.20.5-3
```

m0vie:
Oh, actually this *is* upper-case `X` - so the same thing that exists also on Linux:

It's just that Cygwin is case-insensitive which causes this :)

danielshahaf:
What's the output of `tests/generate.zsh x main x`?

Same question, when the zsh/parameter codepaths are disabled?

m0vie:
> What's the output of `tests/generate.zsh x main x`?

```
$ ./tests/generate.zsh x main x
Set copyright year to 2020?

BUFFER=$'x'

expected_region_highlight=(
  '1 1 command' # x
)
```

> Same question, when the zsh/parameter codepaths are disabled?

Same. (I added `zmodload -u zsh/parameter` at the top of `tests/generate.zsh`).

danielshahaf:
> Same. (I added `zmodload -u zsh/parameter` at the top of `tests/generate.zsh`).

This is not an effective method of disabling those codepaths.

You should have figured that out by yourself, without me needing to point it out for you; for example, by running:

```sh
% git grep -n zsh/parameter
⋮
zsh-syntax-highlighting.zsh:581:zmodload zsh/parameter 2>/dev/null || true
```

m0vie:
Oh, of course, sorry.

I removed that line and explicitly replaced it with `zmodload -u zsh/parameter` to be sure.

The output of `tests/generate.zsh x main x` remains the same, nonetheless.


Anything left to do here in this PR?

Using something more expressive (`global_alias`) instead of a `x` can't hurt either way.

danielshahaf:
Nothing left to do; I just wanted to double-check that case insensitivity was handled correctly while we were at it.  Thanks for verifying that.  Milestone is already set.  Cheers.

PR #776, opened 2020-10-29
Alias fix

### LINKED ISSUE #775: Aliasing problem on zsh 5.8
Hi there,

First of all, thanks for this plugin, using since forever and it is awesome.

I've just upgraded to zsh 5.8 and my terminal stuck in one case. It looks reproducible:

- Have tmux installed
- Have only this block in your .zshrc:
```
if ! [[ "$TERM" =~  '.*screen.*' ]]
then
    # Notify if there is tmux available
    if tmux ls > /dev/null 2>&1; then
        # reminder
        echo "tmux session available"
    fi
    alias tmux="tmux -u att || tmux -u new-session -s default"
fi
```
- Source the plugin.

Afterwards if you want to type `tmux` it gets stuck once you press **x**. If I define 2 aliases without ||, it seems to work, but I didn't understand why it didn't like double pipe especially in this case so wanted to report.

Sorry if I am doing something obviously wrong.
Thx

//edit: I've installed with zplug (with defer:2 as instructed), OS is Linux/NixOS.

phy1729:
Should fix #775. The first commit makes in_alias an array of shift values to keep track of how nested we are, and the second commit makes the value of seen_alias how nested we were when the alias was seen so when a command separator is hit only the aliases of this layer or deeper are forgotten.

Should sleep on this and review tomorrow, but it passes tests.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -601,6 +601,12 @@ _zsh_highlight_main_highlighter_highlight_list()
       (( in_alias[1]-- ))
       # Remove leading 0 entries
       in_alias=($in_alias[$in_alias[(i)<1->],-1])
+      (){
+        local alias_name
+        for alias_name in ${(k)seen_alias[(R)<$#in_alias->]}; do
+          unset "seen_alias[$alias_name]"
+        done
+      }
       if (( $#in_alias == 0 )); then
         seen_alias=()
         # start_pos and end_pos are of the alias (previous $arg) here
@@ -699,7 +705,7 @@ _zsh_highlight_main_highlighter_highlight_list()
           _zsh_highlight_main_add_region_highlight $start_pos $end_pos unknown-token
           continue
         fi
-        seen_alias[$arg]=1
+        seen_alias[$arg]=$#in_alias
         _zsh_highlight_main__resolve_alias $arg
         local -a alias_args
         # Elision is desired in case alias x=''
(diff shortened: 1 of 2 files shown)


highlighters/main/main-highlighter.zsh
```diff
@@ -76,7 +76,7 @@ _zsh_highlight_main_add_region_highlight() {
   integer start=$1 end=$2
   shift 2
 
-  if (( in_alias )); then
+  if (( $#in_alias )); then
```
danielshahaf, line 79: Does `(( ${+in_alias[1]} ))` perform measurably better?

highlighters/main/main-highlighter.zsh
```diff
@@ -596,9 +597,18 @@ _zsh_highlight_main_highlighter_highlight_list()
     last_arg=$arg
     arg=$args[1]
     shift args
-    if (( in_alias )); then
-      (( in_alias-- ))
-      if (( in_alias == 0 )); then
+    if (( $#in_alias )); then
+      (( in_alias[-1]-- ))
```
danielshahaf, line 601: `arrayvar[-1]` has to walk the array (twice, IIRC), so I wonder whether inserting to the array at the start (rather than the end) would perform better.

highlighters/main/main-highlighter.zsh
phy1729: Oddly enough it preforms measurably worse.
```
a=(); time ( repeat 100000; do (( ${+a[1]} )); done; )
```
array | `$#a` | `${+a[1]}`
-- | -- | --
`a=()` | 0.16 | 0.29
`a=(1 2 3)` | 0.17 | 0.32

highlighters/main/main-highlighter.zsh
phy1729: It also fixes the issue with 4.3.11.

highlighters/main/main-highlighter.zsh
danielshahaf: Curious, yes.  Agree that no change is needed in this case, but since we use `${+a[1]}` elsewhere, it's worth noting that that idiom _is_ faster for large arrays.

danielshahaf: Thanks for testing!

PR #788, opened 2020-12-24
Switch to GitHub Actions from Travis

phy1729:
Travis has been running slowly if at all lately. Use GitHub Actions instead which also seems to run faster than Travis did previously.

The schedule cron is simply because, if memory serves, Travis would re-run daily. zsh development rarely moves that fast, so I figure weekly is probably alright. Not tied to the particular time if another is preferred.


phy1729:
The docs for GitHub Actions are at https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions

Example runs are at https://github.com/phy1729/zsh-syntax-highlighting/actions . The IRC trigger did send a message to the IRC channel before I added the extra repository owner condition.

phy1729:
@danielshahaf merged so that tests will be functional again. Can revert or modify if you'd prefer a different course.

danielshahaf:
> @danielshahaf <https://github.com/danielshahaf> merged so that tests 
> will be functional again. Can revert or modify if you'd prefer a 
> different course.

@phy1729 I don't expect to have any reservations; I saw the PR, but I wasn't directly addressed, had no comments to offer, and was happy to go along with whatever you deemed best, so I said nothing.  Thanks for fixing this!

PR #796, opened 2021-02-25
main: Fix alias loops in nested lists (e.g. command substitution)

phy1729:
`alias ls='echo $(ls)'` then `ls` will cause zsh to crash. This avoids that.

The idea is that we are still in the alias even when recursing into a nested list like a command substitution, so _highlight_list is the wrong place to local seen_alias.

### DIFF
--- a/highlighters/main/test-data/alias-loop-commandsubstitution.zsh
+++ b/highlighters/main/test-data/alias-loop-commandsubstitution.zsh
@@ -1,0 +1,37 @@
+#!/usr/bin/env zsh
+# -------------------------------------------------------------------------------------------------
+# Copyright (c) 2021 zsh-syntax-highlighting contributors
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are permitted
+# provided that the following conditions are met:
+#
+#  * Redistributions of source code must retain the above copyright notice, this list of conditions
+#    and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright notice, this list of
+#    conditions and the following disclaimer in the documentation and/or other materials provided
+#    with the distribution.
+#  * Neither the name of the zsh-syntax-highlighting contributors nor the names of its contributors
+#    may be used to endorse or promote products derived from this software without specific prior
+#    written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
+# IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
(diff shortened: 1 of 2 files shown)


highlighters/main/test-data/alias-loop-commandsubstitution.zsh
```diff
@@ -0,0 +1,37 @@
...
+# -------------------------------------------------------------------------------------------------
+# -*- mode: zsh; sh-indentation: 2; indent-tabs-mode: nil; sh-basic-offset: 2; -*-
+# vim: ft=zsh sw=2 ts=2 et
+# -------------------------------------------------------------------------------------------------
+
+alias ls='echo "$(ls)"'
+
+BUFFER='ls'
+
+expected_region_highlight=(
```
danielshahaf, line 35: Mention that this test will crash zsh (infinite recursion, presumably?) if the bug being tested for is present?

danielshahaf:
> The idea is that we are still in the alias even when recursing into a nested list like a command substitution, so _highlight_list is the wrong place to local seen_alias.

What about the opposite problem?  Might some $seen_alias element remain set _longer_ than it should?

phy1729:
I can't imagine a scenario in which an alias is being expanded and in the course of expansion it is again eligible for expansion. That way lies madness^Wstack overflows.

danielshahaf:
The C code enforces that an alias be ineligible for expansion whilst it's being expanded. It's easy to see that: `alias ls=ls` works.

danielshahaf:
I don't immediately follow how that's the same thing as what I asked, though. I'll look again later.

phy1729:
Seems `alias ls='eval ls'` then `ls` will cause zsh to segfault. Of course it's possible to make such a construct not segfault; however, both with and without the patch z-sy-h will highlight the alias as such since we don't peek into eval.

The point being the only way for seen_alias to be set longer than it ought is if an alias is eligible for expansion while it is itself being expanded. That seems to only be the case with `eval`.

danielshahaf:
That answers my question, then. I don't have any other review comments.

Would you report the segfault, please?  Either upstream or here, whichever's appropriate.  Thanks!

danielshahaf:
I don't get a segfault.  Both with and without z-sy-h, I get this:

% alias ls='eval ls'
% ls
zsh: job table full or recursion limit exceeded

I did press <Enter> at the end of the second line.

danielshahaf:
> I don't get a segfault. Both with and without z-sy-h, I get this:
> 
> % alias ls='eval ls'
> % ls
> zsh: job table full or recursion limit exceeded

With PRINT_EXIT_VALUE in effect too, I see exactly 999 copies of "zsh: exit 1" printed.

That looks like it might be MAX_MAXJOBS: https://github.com/zsh-users/zsh/blob/5ede2c55f144593c16498c3131a76e188114a9c6/Src/jobs.c#L2061-L2066

danielshahaf:
Should this be milestoned 0.8.0?

PR #808, opened 2021-04-10
Fixed broken fish shell link in README.md

robert5800:
The correct link should be https://fishshell.com/


danielshahaf:
Thanks, @robert5800.  Changed the log message, rebased, and merged in 993a07fc7f4e077b556192cfc5c3efc29b40fd96.

Would you report this to https://github.com/fish-shell/ as well, please?  Looks like `www.fish-shell.com.` and `fish-shell.com.` resolve to different IP addresses.

Issue #823, opened 2021-07-26
All custom aliases removed


bagnaram:
https://github.com/zsh-users/zsh-syntax-highlighting/blob/dffe304567c86f06bf1be0fce200077504e79783/zsh-syntax-highlighting.zsh#L36

This line appears to be resetting all custom aliases I have set on my system. This is problematic because I enable/disable this plugin asynchronously doring my terminal sessions

This command should be something like `unalias -m '^+.*'`

bagnaram:
Created https://github.com/zsh-users/zsh-syntax-highlighting/pull/824

phy1729:
Aliases are (normally) restored later
https://github.com/zsh-users/zsh-syntax-highlighting/blob/dffe304567c86f06bf1be0fce200077504e79783/zsh-syntax-highlighting.zsh#L593

Most aliases can make the round trip unchanged. Aliases starting with `+` are subject to a bug in some version of zsh, so are excluded from being unset. Perhaps see if any of your aliases are using odd characters which have a similar issue.

bagnaram:
I figured out that the issue is that the alias is being lost by my package manager, zinit.

bagnaram:
https://github.com/zdharma/zinit/issues/530 for  reference

PR #824, opened 2021-07-26
unalias fix

bagnaram:
fix regex to not match all aliases. Currintly it blows away all custom alias


phy1729:
Change not needed (cf. #823).

Issue #829, opened 2021-08-15
Trigger highlighting form a script


typkrft:
This may be an odd scenario, but I've got a script (an fzf powered cheatsheet) that writes output to the prompt using print -z eg: `print -z -- $(printf "printf \"hello, world\"")`. Unfortunately when it prints to the command line the output does not have any highlighting, moving the cursor will trigger it, but I was wonder if there is another way something automatic. I've tried emulating cursor movement with tput to no avail. It's not a huge deal, but I thought I'd ask. Thanks!

danielshahaf:
Looks like this behaves the way you want on the redrawhook codepath.  Try enabling it manually by editing the condition in the source, here:

https://github.com/zsh-users/zsh-syntax-highlighting/blob/dffe304567c86f06bf1be0fce200077504e79783/zsh-syntax-highlighting.zsh#L400-L417

typkrft:
I'm sorry, but I'm not really understanding what you mean. What should I try to enable?

danielshahaf:
> I'm sorry, but I'm not really understanding what you mean. What should 
> I try to enable?

At the line I linked (417), change "5.8.0.2" to "5.3".

What's the output of `typeset -p ZSH_VERSION ZSH_PATCHLEVEL`?

typkrft:
That did the trick. Thank you!

$ `typeset -p ZSH_VERSION ZSH_PATCHLEVEL`
```Shell
typeset ZSH_VERSION=5.8
typeset ZSH_PATCHLEVEL=zsh-5.8-0-g77d203f
```

danielshahaf:
You're welcome. That codepath is on by default when running under the next zsh release, so it seems there's nothing else to do here.

Issue #513, opened 2018-04-05
Works so slow when I try to paste long string to terminal


haoranpb:
I am just using `main` highlighter. At first, I thought if I just disable some styles, it will work fine. So I disabled all `quoted-argument`. 

When I paste `magnet:?xt=..t.nz` after `"` into the terminal:

`aria2c "magnet:?xt=urn:btih:90...er.fastcast.nz"` 

The highlight is gone, but it's still really slow!
I really love the highlighting, how can I make it faster?

phy1729:
(Related to #240.)

Being slow with large inputs is a known issue. There seems to be two main causes:
1) Iteration over a string is quadratic.
2) Array appends are quadratic.

I've briefly looked at solving the first in zsh, but I haven't had time to make a complete patch.

With regards to disabling styles, setting a style to none doesn't disable the code that parses for that style. We could add a flag to disable `_zsh_highlight_main_highlighter_highlight_argument`, but that would disable
* single-hyphen-option
* double-hyphen-option
* history-expansion (sometimes)
* globbing
* path (and related styles)
* single-quoted-argument (and -unclosed)
* rc-quote
* double-quoted-argument (and -unclosed)
* dollar-double-quoted-argument
* back-double-quoted-argument
* dollar-quoted-argument (and -unclosed)
* back-dollar-quoted-argument
* back-quoted-argument (and -unclosed)

and once #512 is merged the highlighting of command and process substitution and their contents.

If that's acceptable, you can add `return 0` as the first command in `_zsh_highlight_main_highlighter_highlight_argument`. We could add a flag to do the same, but I think long term it'd be preferable to fix the quadratic issues in zsh.

ludanxer:
Thx a lot!

 I didn't really find a func named `_zsh_highlight_main_highlighter_highlight_argument`, but I think you mean add `return 0` in front of [line 652 in main-highlighter.zsh](https://github.com/zsh-users/zsh-syntax-highlighting/blob/e4fffa4de566c9552f9284fd75e689edb3a1083d/highlighters/main/main-highlighter.zsh#L652)

It did work a little faster

Tip in #86 is a good idea, I think I will just disable highlighting when I have a very large input

phy1729:
Ah forgot the current release doesn't have that function. Where you added it will do similar, but may mess up the state in some cases. Don't think there's a single line change that'd work in the current release.

[...]

PR #835, opened 2021-08-29
skip command line parsing when there are still keys in the input queue

dkuettel:
If you are working with paths that are on a slow filesystem (eg, NFS shares), then highlighting can become sluggish at times. 

`ZSH_HIGHLIGHT_DIRS_BLACKLIST+=(/some/path /and/more/paths)` does work but there is still somewhere a small slow down, I couldn't find where exactly it becomes slower when it has to handle paths from a slow filesystem compared to a fast filesystem. 

The slow down is so small that you dont notice a problem while typing. But when you paste in a 60-character path, it blocks for a while, because in that case it looks like 60 key events to zsh and it causes a highlighter pass everytime. The "small slow down times 60" is then very apparent.

One solution might be to skip parsing when there is still data in the input queue (see `KEYS_QUEUED_COUNT` in a zsh zle widget). That's typically the case with pasting. Skipping parsing seems okay to me in that case since you know you will get called immediately again because there are more keys to be processed.

Maybe this can be useful, or an approach inspired by it. It doesnt have to be this exact commit.

### DIFF
--- a/zsh-syntax-highlighting.zsh
+++ b/zsh-syntax-highlighting.zsh
@@ -219,7 +219,8 @@ _zsh_highlight()
   [[ -n ${ZSH_HIGHLIGHT_MAXLENGTH:-} ]] && [[ $#BUFFER -gt $ZSH_HIGHLIGHT_MAXLENGTH ]] && return $ret
 
   # Do not highlight if there are pending inputs (copy/paste).
-  [[ $PENDING -gt 0 ]] && return $ret
+  (( KEYS_QUEUED_COUNT > 0 )) && return $ret
+  (( PENDING > 0 )) && return $ret
 
   {
     local cache_place


phy1729:
This seems like a good idea; though I think it would be better to put the check in `_zsh_highlight_highlighter_main_predicate` or `_zsh_highlight_buffer_modified`, so all highlighters benefit from the same logic. I have a slight preference for the latter, but could be convinced if someone has a strong opinion the other way.

(Related issue #513)

danielshahaf:
Agreed.

Wondering whether this is needed in the redrawhook codepath (https://github.com/zsh-users/zsh-syntax-highlighting/issues/829#issuecomment-899971930).

dkuettel:
> This seems like a good idea; though I think it would be better to put the check in `_zsh_highlight_highlighter_main_predicate` or `_zsh_highlight_buffer_modified`, so all highlighters benefit from the same logic. I have a slight preference for the latter, but could be convinced if someone has a strong opinion the other way.
> 
> (Related issue #513)

I dont have a very good overview of the full code. My change here was quite opportunistic. If we implement it where you said then it only applies for the `main` highlighter. Would it make sense to go up one more level? Then it would apply to any highlighter, generically.

As for which of the two functions, I have little stake in the choice, but I like to point out that for `_zsh_highlight_highlighter_main_predicate` the comment says "Whether the highlighter should be called or not.", whereas for `_zsh_highlight_buffer_modified` it says "Whether the command line buffer has been modified or not.". The first one fits the description better. In effect, both places work.

If there is no other opinion, I'll put it where phy1729 said.

What do you think about moving it up one more level to make it work for any highlighter?

danielshahaf:
`_zsh_highlight_buffer_modified` is used by multiple highlighters. Note its name doesn't include a `_main_` namespace component.

We could make the new condition a condition for calling the …_predicate function at all, or make it a function that multiple highlughters' `…_predicate` functions call. WDYT?

dkuettel:
Ah yes I see why you said `_zsh_highlight_buffer_modified` is the better choice. I moved it and it works well for my local use case. I'm fine with this approach. I did change the comment a bit to make it clear.

A new condition might be more "beautiful", but it would also mean we need to make sure that all other parser now call it, and there are probably 3rd party highlighters that we cannot control. My vote is to leave it as it is now.

If we agree, we can move on to make the tests pass :) I guess they fail because they expect the highlighting to be applied between keystrokes?

dkuettel:
(I actually force pushed the change, apologies if that is not this repo's policy.)

danielshahaf:
> A new condition might be more "beautiful", but it would also mean we need to make sure that all other parser now call it, and there are probably 3rd party highlighters that we cannot control. My vote is to leave it as it is now.

Disagree. It's perfectly fine to add new API's in new releases. Conversely, we can't never add API's just because someone might be using existing API's.

If anything, having `_zsh_highlight_buffer_modified` do another check would break its API contract for its existing callers, as well as violate the principle of 'do one thing and do it well'. (On the other hand, if we had a 'do some default, sane pre-flight checks' function for `*_predicate` functions to use, extending it would likely have been reasonable.)

> If we agree, we can move on to make the tests pass :) I guess they fail because they expect the highlighting to be applied between keystrokes?

Could you post the output of `make quiet-test`? I don't see it in github's UI.

Tests highlight once at the end; see how they fake the value of $CURSOR (tests/test-highlighting.zsh:154).

> (I actually force pushed the change, apologies if that is not this repo's policy.)

Feel free to do whatever's easiest for your reviewers and you.

dkuettel:
Ok then, here is another draft. I made a new utility `_zsh_highlight_input_queue_empty` and call it in all `predicate`s.

For me, `make quiet-test` runs without errors.

https://github.com/zsh-users/zsh-syntax-highlighting/blob/master/docs/highlighters.md#how-to-implement-a-new-highlighter doesnt mention `_zsh_highlight_buffer_modified` or similar things currently. When someone implements a new highlighter they would probably take inspiration from the existing once. Do you want any changes in the documentation?

zsh-syntax-highlighting.zsh
```diff
@@ -358,6 +358,16 @@ _zsh_highlight_buffer_modified()
...
 }
 
+# Whether zsh is done processing the input queue.
+# Note that for many highlighters there is no need parse the
+# current buffer state if there are still keys to be processed.
+#
+# Returns 0 if there are no more keys to be processed, ie, the buffer likely won't change for a while.
+_zsh_highlight_input_queue_empty()
+{
+    [[ ${KEYS_QUEUED_COUNT:-0} == 0 ]]
```
phy1729, line 368: I presume the `:-0` is so the tests run under `WARN_CREATE_GLOBAL`. Since `KEYS_QUEUED_COUNT` is a ZLE variable, I think it'd make sense to spoof it in the test harness (tests/test-highlighting.zsh in the `run_test_internal` function) along with `CURSOR` and `PENDING`.

Since the variable is numeric, you could use an arithmetic expression (e.g. `(( KEYS_QUEUED_COUNT == 0 ))`), but that's just style.

phy1729 (changes requested):
I'm not sure if `KEYS_QUEUED_PENDING` should overrule `[[ $WIDGET == zle-line-finish ]]` or the other way around. Thoughts?

danielshahaf:
The line-finish check is there to drop path_prefix highlighting of the last word when that's followed by accept-line; that should still be done even if the queue is non-empty. So, I could be wrong, but shouldn't it be:

> line-finish || (queue empty && modified)

?  I'm proposing this as the semantics (truth table) to implement only; i.e., I haven't tried to optimize the short-circuiting behaviour.

zsh-syntax-highlighting.zsh
dkuettel: Ah interesting, I just used `${KEYS_QUEUED_COUNT:-0}` for "good measure" in case it's not set in some older version or for some other reason. With that default it should always behave reasonably. But it looks like I circumvented testing a bit with this.

It makes sense to spoof it. Should we rely on the variable to exist? I'm not sure if `zsh-syntax-highlighting` is using `set -u` overerall.

dkuettel:
As I tried to address the review, looking at some of the predicates that I adapted in a naive way, like, eg,

```
# Whether the brackets highlighter should be called or not.
_zsh_highlight_highlighter_brackets_predicate()
{
  _zsh_highlight_input_queue_empty && {
    [[ $WIDGET == zle-line-finish ]] || _zsh_highlight_cursor_moved || _zsh_highlight_buffer_modified
  }
}
```
apart from `zle-line-finish` vs `queue-empty` I'm wondering now if my change is actually robust:

Let's say it's a simpler predicate
`_zsh_highlight_input_queue_empty && _zsh_highlight_buffer_modified`
which used to be just
`_zsh_highlight_buffer_modified`
now what if the input queue contains "ls test←←" (two left arrow keys).
It will hold off doing anything until it has processed all of it.
The last two operations only change the cursor position.
But they dont change the buffer contents.
Will it then conclude that nothing has changed and it wont run highlighting?

```
# Whether the command line buffer has been modified or not.
#
# Returns 0 if the buffer has changed since _zsh_highlight was last called.
_zsh_highlight_buffer_modified()
{
  [[ "${_ZSH_HIGHLIGHT_PRIOR_BUFFER:-}" != "$BUFFER" ]]
}
```

This probably compares after the last "←" to the buffer contents after "ls test←", right?

phy1729:
While testing your example, I discovered we already test for if `PENDING` is non-zero and bail out early if so. https://github.com/zsh-users/zsh-syntax-highlighting/blob/6e0e950154a4c6983d9e077ed052298ad9126144/zsh-syntax-highlighting.zsh#L221-L222 I think that means this PR would not change anything and can be dropped. Do you agree?

dkuettel:
> we already test for if `PENDING`

Curious, my original problem persists, so somehow that doesnt seem to work, on my system at least.

`man zsh` says

```
         KEYS_QUEUED_COUNT (integer)
                The number of bytes pushed back to the input queue and therefore available for reading immediately before any I/O
                is done; read-only.  See also PENDING; the two values are distinct.
```

and

```
        PENDING (integer)
                The number of bytes pending for input, i.e. the number of bytes which have already been typed and can immediately
                be read. On systems where the shell is not able to get this information, this parameter will always have a value
                of zero.  Read-only.  See also KEYS_QUEUED_COUNT; the two values are distinct.
```

potentially `KEYS_QUEUED_COUNT` is more reliable? I'm not sure if I should believe that sometimes the system might know the key count in the queue but not the byte count.

I can try to find out why in my setting your referenced code doesn't seem to skip highlighting.

The other problem I described, that it might skip a few times because `PENDING` and then ultimately not do anything because `BUFFER` didn't change. Do you know if that does not apply here?

dkuettel:
From my tests on zsh `zsh 5.8 (x86_64-ubuntu-linux-gnu)` indeed when I paste a longer path and add

```echo "PENDING=$PENDING KEYS_QUEUED_COUNT=$KEYS_QUEUED_COUNT" >> ~/hl-log```

just before the test mentioned above. I get

```
PENDING=0 KEYS_QUEUED_COUNT=78
PENDING=0 KEYS_QUEUED_COUNT=77
...
PENDING=0 KEYS_QUEUED_COUNT=4
PENDING=0 KEYS_QUEUED_COUNT=3
PENDING=0 KEYS_QUEUED_COUNT=2
PENDING=0 KEYS_QUEUED_COUNT=1
PENDING=0 KEYS_QUEUED_COUNT=0
```

When I use `[[ $KEYS_QUEUED_COUNT -gt 0 ]]` instead of `[[ $PENDING -gt 0 ]]` it works well with my original problem. As far as this PR is concerned I think I would make this small change then.

I still wonder what you think about skipping because `KEYS_QUEUED_COUNT` and then at the end no highlighting because the very last key was not a buffer-changing operation. The code around `zsh-syntax-highlighting/zsh-syntax-highlighting.zsh:222` is very involved. Do you have any insight?

danielshahaf:
> I'm not sure if I should believe that sometimes the system might know the key count in the queue but not the byte count.

Actually, the docs you quoted explicitly say that both parameters' values count bytes.  The important difference is the one between "pushed back" and "pending".  The former implies some sort of `ungetc()`-like functionality.

Let's see:

https://github.com/zsh-users/zsh/blob/dd51ffa5b4b9759af2667df8e4505f117b8e2b23/Src/Zle/zle_params.c#L525-L530
https://github.com/zsh-users/zsh/blob/dd51ffa5b4b9759af2667df8e4505f117b8e2b23/Src/utils.c#L2996-L3016

https://github.com/zsh-users/zsh/blob/dd51ffa5b4b9759af2667df8e4505f117b8e2b23/Src/Zle/zle_params.c#L467-L472
https://github.com/zsh-users/zsh/blob/dd51ffa5b4b9759af2667df8e4505f117b8e2b23/Src/Zle/zle_main.c#L188

$KEYS_QUEUED_COUNT returns the length of an internal `ungetc()`-like buffer while $PENDING does a syscall asking the tty what's queued.  The "doesn't know" part refers to systems that don't have `FIONREAD`, used by $PENDING's getfn's callee.

tl;dr: The documentation is correct.

> When I use `[[ $KEYS_QUEUED_COUNT -gt 0 ]]` instead of `[[ $PENDING -gt 0 ]]` it works well with my original problem. As far as this PR is concerned I think I would make this small change then.

Your analysis here isn't valid.  The fact that you couldn't reproduce a situation in which $PENDING is non-zero and $KEYS_QUEUED_COUNT is zero is outweighed by the explicit warnings in the documentation that $PENDING will always be zero on some systems and that the two parameters are "distinct" (meaning, I presume, that neither of them implies the value of the other).

Having reviewed the implementation, it is in fact possible for $PENDING to be non-zero and $KEYS_QUEUED_COUNT to be zero simultaneously.  It should be easy to reproduce this in the debugger.

Support for $PENDING was added in commit 0999ab05bafe034fdcbca507f369b669a0b41d35.

So, my guess would be that both parameters should be tested ($KEYS_QUEUED_COUNT first, to take advantage of short-circuiting).

> I still wonder what you think about skipping because `KEYS_QUEUED_COUNT` and then at the end no highlighting because the very last key was not a buffer-changing operation. The code around `zsh-syntax-highlighting/zsh-syntax-highlighting.zsh:222` is very involved. Do you have any insight?

First, let me link to [line 222](https://github.com/zsh-users/zsh-syntax-highlighting/blob/6e0e950154a4c6983d9e077ed052298ad9126144/zsh-syntax-highlighting.zsh#L222).

I don't see the problem.  At the time of the very last key, `_zsh_highlighting_buffer_modified` compares the then-current buffer not to the buffer at the time of the penultimate key but to the buffer as of the last time highlighting was performed.  All highlighters that call this function are "memoryless" — that is, their output depends only on the current values of $PREBUFFER and $BUFFER, as opposed to the sequence of edit operations that brought about those values (contrast [the manual handling of `zle_highlight` special contexts](https://github.com/zsh-users/zsh-syntax-highlighting/blob/6e0e950154a4c6983d9e077ed052298ad9126144/zsh-syntax-highlighting.zsh#L263-L293) and [_zsh_highlight_highlighter_cursor_predicate](https://github.com/zsh-users/zsh-syntax-highlighting/blob/6e0e950154a4c6983d9e077ed052298ad9126144/highlighters/cursor/cursor-highlighter.zsh#L35-L39)) — so I don't see a problem.

I suppose you could break things if you had, say, a zle widget that created or deleted aliases, and invoked it in such a way that the invoking keypress was skipped by the KEYS_QUEUED_COUNT/PENDING mechanism, thereby causing the created/deleted aliases to not change their highlighting correctly.

dkuettel:
> Actually, the docs you quoted explicitly say that both parameters' values count bytes. The important difference is the one between "pushed back" and "pending". The former implies some sort of `ungetc()`-like functionality.

Thanks for the patient explanation. You're right. I didnt read it carefully in the man-page.

> So, my guess would be that both parameters should be tested ($KEYS_QUEUED_COUNT first, to take advantage of short-circuiting).

I agree. In a way we want `(( PENDING + KEYS_QUEUED_COUNT > 0 ))` but we can short-circuit it.

> I don't see the problem.

Good then :)

dkuettel:
`make quiet-test` succeeded for me

dkuettel:
Is there something I didnt address yet?

danielshahaf:
Seems ready to merge to me.  Matthew, WDYT?  @dkuettel, sorry for leaving this hanging; the maintainers are short on tuits.

phy1729 (approved):

phy1729:
Apologies for the lag. Looks good to me as well. Merged. Thank you @dkuettel for PR!

ICHx:
new to zsh-syntax-highligthing, wanted to say that this PR let me changed from using `zsh-fast-syntax-highlighting` to here.

PR #836, opened 2021-09-10
Added `regexp` to the list of built-in highlighters in documentation

leogama:
Added the link to the `regexp` highlighter documentation file and specified that the `pattern` highlighter uses globbing to make the difference between the two explicit.

### DIFF
--- a/docs/highlighters.md
+++ b/docs/highlighters.md
@@ -6,6 +6,7 @@ Syntax highlighting is done by pluggable highlighters:
 * `main` - the base highlighter, and the only one [active by default][main].
 * `brackets` - [matches brackets][brackets] and parenthesis.
 * `pattern` - matches [user-defined patterns][pattern].
+* `regexp` - matches [user-defined regular expressions][regexp].
 * `cursor` - matches [the cursor position][cursor].
 * `root` - highlights the whole command line [if the current user is root][root].
 * `line` - applied to [the whole command line][line].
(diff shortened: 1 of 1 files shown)


phy1729:
Thanks for spotting this and sending a PR! Having to renumber for each new highlighter is non-ideal. Would you mind making a commit that just changes the numbers to labels (similar to INSTALL.md) and then a second commit that adds regexp to the list of highlighters? If you'd rather not, I could do that while crediting you in the commit message.

docs/highlighters.md
```diff
@@ -3,19 +3,19 @@ zsh-syntax-highlighting / highlighters
 
 Syntax highlighting is done by pluggable highlighters:
 
-* `main` - the base highlighter, and the only one [active by default][1].
-* `brackets` - [matches brackets][2] and parenthesis.
```
phy1729, line 7: Looks like a space was lost here.

phy1729:
Just a small nit generally commit messages should be in the imperative (e.g. `docs: Change highlighters' URL` and `docs: Add regexp`), but that doesn't block merging. Other than the lost space, looks good to merge.

docs/highlighters.md
leogama: Good sight! I almost missed it again...

leogama:
Finally! Ready to go for merging if everything is right.

danielshahaf:
Merged, thanks!

While reviewing this I noticed that we don't seem to explain anywhere, neither in this file nor in docs/highlighters/pattern.md, what flavour of patterns that highlighter uses (i.e., https://zsh.sourceforge.io/Doc/Release/Expansion.html#Filename-Generation).  That issue was preëxisting and beyond the scope of the PR, so didn't block merging.

PR #839, opened 2021-10-13
Organize main highlighters alphabetically

Twoody:
For better readability and usability (seeing all possible options), alphabetical sorting is easier for users.


phy1729:
> Instead, we could group the entries semantically

I think the question is if the docs are intended more for reference or pedagogy. For the former I would strongly argue alphabetical is best. For the latter grouping semantically would be better (perhaps with different levels of bullet points to denote fallback?). The rest of the file doesn't feel like a reference; so while I am fond of alphabetized lists, I think @danielshahaf is correct. Perhaps something like

danielshahaf:
> I think the question is if the docs are intended more for reference or pedagogy.

I assume that by "reference" you mean a random-access use case ("What's the `foo` style?") and by "pedagogy" you mean more of a read-through use case ("What features exist?")?

> Not sure what to call things like assign, redirection, comments and the like.

"Miscellany"?  "Other syntactical elements"?

PR #845, opened 2021-11-11
Emphasize configuration location for oh-my-zsh installation instructions

benclauss:
See https://github.com/spf13/cobra/issues/1534 for motivation.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -83,11 +83,13 @@ your `.zshrc`.
     git clone https://github.com/zsh-users/zsh-syntax-highlighting.git ${ZSH_CUSTOM:-~/.oh-my-zsh/custom}/plugins/zsh-syntax-highlighting
     ```
 
-2. Activate the plugin in `~/.zshrc`:
+2. Activate the plugin **at the bottom** of `~/.zshrc`:
 
     ```zsh
     plugins=( [plugins...] zsh-syntax-highlighting)
     ```
+
+See the [FAQ](https://github.com/zsh-users/zsh-syntax-highlighting#why-must-zsh-syntax-highlightingzsh-be-sourced-at-the-end-of-the-zshrc-file) for more details.
 
 3. Restart zsh (such as by opening a new instance of your terminal emulator).
 


INSTALL.md
```diff
@@ -83,12 +83,14 @@ your `.zshrc`.
     git clone https://github.com/zsh-users/zsh-syntax-highlighting.git ${ZSH_CUSTOM:-~/.oh-my-zsh/custom}/plugins/zsh-syntax-highlighting
     ```
 
-2. Activate the plugin in `~/.zshrc`:
+2. Activate the plugin **at the bottom** of `~/.zshrc`:
```
danielshahaf, line 86: This could be understood to mean "Append `plugins=( "${plugins[@]}" zsh-syntax-highlighting )` to your zshrc".

I think the right instruction would be to ensure that zsh-syntax-highlighting is last in the array and that the command that reads the array is run at the end of zshrc, but none of the maintainers uses omz so we rely on patch submissions in this regard.

INSTALL.md
```diff
@@ -83,12 +83,14 @@ your `.zshrc`.
...
 
     ```zsh
     plugins=( [plugins...] zsh-syntax-highlighting)
     ```
 
+See the [FAQ](https://github.com/zsh-users/zsh-syntax-highlighting#why-must-zsh-syntax-highlightingzsh-be-sourced-at-the-end-of-the-zshrc-file) for more details.
```
danielshahaf, line 92: Is it possible to use a relative URL here? The documentation gets installed and should be usable offline.

danielshahaf (commented):
See below.

INSTALL.md
```diff
@@ -83,12 +83,14 @@ your `.zshrc`.
     git clone https://github.com/zsh-users/zsh-syntax-highlighting.git ${ZSH_CUSTOM:-~/.oh-my-zsh/custom}/plugins/zsh-syntax-highlighting
     ```
 
-2. Activate the plugin in `~/.zshrc`:
+2. Activate the plugin **at the bottom** of `~/.zshrc`:
 
     ```zsh
     plugins=( [plugins...] zsh-syntax-highlighting)
```
saheljalal, line 89: This doesn't work for me. I had to add spaces like this or it errored out:
```sh
plugins=( [ plugins... ] zsh-syntax-highlighting )
```

danielshahaf: Those brackets are not meant to be entered literally.  Wherever you entered that command literally with spaces, you just disabled all your other plugins.  Change the command to `plugins+=( zsh-syntax-highlighting )`.

This change can and should be made in INSTALL.md as well.

gavindsouza:
I think https://github.com/zsh-users/zsh-syntax-highlighting/pull/870 supersedes this; it makes it easier to update config without manually updating the `.zshrc` file.

PR #846, opened 2021-11-12
Fix typo

rex4539:


phy1729:
I think the text is correct as is (though admittedly therefor is somewhat archaic). I don't really have an opinion, but would default to leave things as is. @danielshahaf ?

OT: Since you appear to be on a typo fixing spree, is there some linter you're using to find these typos? Might be handy to put into some CI pipelines.

danielshahaf:
Yes, it's correct as is.  I'm happy to consider alternatives that improve clarity, of course.  I was trying to describe a set of circumstances first and the behaviour ~~therein~~ in those circumstances second, and used _therefor_ so as to not have to name the function the sentence talks about.  We could reverse that last criterion:

```wdiff
# will succeed.  In particular, if [-a function-]{+F+} has been marked for autoloading
# but is not available in $fpath, then {+calling+} this function {+on F+} will return [-False therefor.-]{+False.+}
```

danielshahaf:
I'll reopen this later pending decision on the proposed text in my last comment. And we did forget to thank you for the patch, so, thanks for the patch :-)

PR #855, opened 2022-02-06
Shallow git cloning

AlfredJKwack:
When installing either manually or with oh-my-zsh it's better to create a shallow clone with a history truncated to the latest commit only. The installation will be faster and less resource intensive. End users don't care for the entire repo's history and branches anyway.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -38,7 +38,7 @@ See also [repology's cross-distro index](https://repology.org/metapackage/zsh-sy
 Simply clone this repository and source the script:
 
 ```zsh
-git clone https://github.com/zsh-users/zsh-syntax-highlighting.git
+git clone --depth 1 https://github.com/zsh-users/zsh-syntax-highlighting.git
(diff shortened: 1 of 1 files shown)


phy1729:
GitHub discourages the use of shallow clones especially when the clone will be updated in the future [1], so I'm hesitant to recommend shallow clones to users.

A full clone took 3.1s and takes 3.5MiB.
A blobless clone took 4.4s and takes 2.8MiB.
A treeless clone took 4.2s and takes 2.4MiB.

So changing from a full clone is slower. Does take less disk space, but has a greater server load it seems from the blog post. I don't know if it's worth changing.

[1]: https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/

AlfredJKwack:
Hi @phy1729, thanks for the constructive commentary. After reading through [the article](https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/) it seems indeed clear that for the size and nature of the repo the costs will outweigh the benefits. 

Closing with comment.

PR #884, opened 2022-07-08
Update INSTALL.md

WilliamStar007:
Made the document more structured and easier to read.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,65 +1,9 @@
 # Installation
 
+* [Manual (preferred)](#manual-git-clone)
 * [Packages](#packages)
-* [Antigen](#antigen)
+* [Plugin Managers](#plugin-managers)
 * [Oh My Zsh](#oh-my-zsh)
-* [Manual](#manual-git-clone)
-* [Other](#other)
-
-## Packages
-
-| System  | Package |
-| ------------- | ------------- |
-| Ubuntu | [zsh-syntax-highlighting OBS repository][obs-repository], [zsh-syntax-highlighting in Xenial][ubuntu-package] |
-| Debian | [zsh-syntax-highlighting OBS repository][obs-repository], [zsh-syntax-highlighting in Stretch][debian-package] |
-| Fedora / CentOS / RHEL / Scientific Linux | [zsh-syntax-highlighting OBS repository][obs-repository], [zsh-syntax-highlighting in Fedora 24+][fedora-package-alt] |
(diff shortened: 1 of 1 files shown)


GH-SecurityTest-1:
I kept all the links in the original installation guide and followed the general structure in `zsh-autosuggestions` [Install.md](https://github.com/zsh-users/zsh-autosuggestions/blob/master/INSTALL.md). This PR would make the installation guide more uniform across different zsh plugins.

WilliamStar007:
I kept all the links in the original installation guide and followed the general structure of `zsh-autosuggestions` [Install.md](https://github.com/zsh-users/zsh-autosuggestions/blob/master/INSTALL.md). This PR would make the installation guides more uniform across different zsh plugins/packages.

GH-SecurityTest-1:
I see your point. I included the `in stretch` and `in Xenial` links out of respect for the original installation guide, since I am new to this project and do not have much knowledge about these links. It's just that I worked with both `zsh-autosuggestions` and `zsh-syntax-highlighting`, and their installation guide was much easier to read and follow. So, I thought updating it would be a good idea. That's all. With that said, I think it would be best to keep them there for now, since they still work, and I do not know if any replacement link exist.

As for alphabetical order, I totally agree. I will also update `Max OS X` to `macOS`, thanks for letting me know!

WilliamStar007:
I wanted to update the Install.md because I worked with both `zsh-autosuggestions` and `zsh-syntax-highlighting`, and I think their installation guide is more organized and easier to follow. Thus, I thought updating this project's install.md would be a good idea. That's all, and I am very open to ideas.

For the `in stretch` and `in Xenial` links, I included them out of respect for the original installation guide since I am new to this project. I have no idea if there are better links to replace them, and thus I think keeping them, for now, is probably a good idea. As for alphabetical order and changing `Max OS X` to `macOS`, I totally agree. I will update this PR. Thanks for the advice!

WilliamStar007:
Changed "Mac OS" to "macOS";
All packages are now in alphabetical order;
Manual Installation is at the top of the guide since it is preferred;
`Oh My ZSH` is still a separate section because I think it is very popular. (just my personal opinion)

phy1729:
Special casing Oh My Zsh is giving it special treatment which isn't ok since z-sy-h should remain neutral. Aside from the changes mentioned in my previous comment, I think the page should remain basically as is.

WilliamStar007:
ok... I removed the special case for Oh My Zsh.
I think that my proposed organization and structure are better than the current one (that's why I made a pull request). However, it is indeed a difference in opinion with no objective measurements. So, I guess it is up to the maintainers.

I really like this project and am currently using it. I wish this project all the best.

PR #908, opened 2022-12-14
Add a missing "$" on a variable usage

### LINKED ISSUE #907: sed -e missing command
## How to reproduce

Trying to commit with the result of a `date` command as the message:

1. Typing: `git commit -m "$(d`

## Current failure:

image

Each letter I type is followed by the message:

> sed: -e expression #1, char 1: missing command

I think missing the color in the nested shell would be ok, but the current warning on each letter is quite inconvenient.

Thanks for this project BTW :)

CharlesGueunet:
Fixes #907

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -1537,7 +1537,7 @@ _zsh_highlight_main_highlighter_highlight_double_quote()
               (( i += 2 ))
               _zsh_highlight_main_highlighter_highlight_list $(( start_pos + i - 1 )) S $has_end $arg[i,-1]
               ret=$?
-              (( i += REPLY ))
+              (( i += $REPLY ))
               last_break=$(( start_pos + i ))
               reply=(
                 $saved_reply


CharlesGueunet:
@danielshahaf please review

highlighters/main/main-highlighter.zsh
```diff
@@ -1537,7 +1537,7 @@ _zsh_highlight_main_highlighter_highlight_double_quote()
               (( i += 2 ))
               _zsh_highlight_main_highlighter_highlight_list $(( start_pos + i - 1 )) S $has_end $arg[i,-1]
               ret=$?
-              (( i += REPLY ))
+              (( i += $REPLY ))
```
danielshahaf, line 1540: Thanks for the patch.  Unfortunately, it's not a complete fix.

Arithmetic expressions are parsed for substitutions and then, in a second pass, evaluate identifiers into their values.  Thus, the incumbent code is correct, and even idiomatic (consider `foo=2+2; echo $((foo*2)) $(($foo*2))`), and conversely, the patch would seem to merely paper over the issue.

I'm curious why it works, but we can discuss that back on the issue's ticket.

PR #914, opened 2023-01-31
Fix proxychains args

DiegoBoy:
Args for proxychains should be f:q => -f file  : -q quiet.

From comments in the same function:
``` 
# $precommand_options maps precommand name to values of $flags_with_argument, 
# $flags_sans_argument, and flags_solo for that precommand, joined by a 
# colon. 
```


danielshahaf:
Thanks, merged.  And for anyone reading this in the future, please include a link to the tool's upstream man page so it's easier for us to review, thanks.  (In this case I went by https://github.com/haad/proxychains/blob/master/src/proxychains.)

PR #920, opened 2023-03-29
fix comments being invisible in certain conditions

### LINKED ISSUE #510: Comments are invisible
This is a commonly used technique: https://unix.stackexchange.com/questions/33994/zsh-interpret-ignore-commands-beginning-with-as-comments?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

But with zsh-syntax-highlighting turned on, all characters after `#` become invisible.

Silejonu:
When `setopt interactivecomments` is set, comments become invisible in a black terminal. This commit fixes it by adding a magenta background.

Of course the choice of colour is debatable, I went with what made the most sense to me.

`fg=black` could also be removed entirely, so that the text is in the same colours as the default text.

Fixes #510

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -59,7 +59,7 @@
 : ${ZSH_HIGHLIGHT_STYLES[back-dollar-quoted-argument]:=fg=cyan}
 : ${ZSH_HIGHLIGHT_STYLES[assign]:=none}
 : ${ZSH_HIGHLIGHT_STYLES[redirection]:=fg=yellow}
-: ${ZSH_HIGHLIGHT_STYLES[comment]:=fg=black,bold}
+: ${ZSH_HIGHLIGHT_STYLES[comment]:=fg=black,bg=magenta,bold}
 : ${ZSH_HIGHLIGHT_STYLES[named-fd]:=none}
 : ${ZSH_HIGHLIGHT_STYLES[numeric-fd]:=none}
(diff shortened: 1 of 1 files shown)


danielshahaf:
Making z-sy-h not use "bold black" doesn't seem like the right fix to me, anyway, since any other application may use "bold black" as well.  I'd sooner recommend to configure your terminal emulator so as many colour combinations as possible are readable.

Indeed, "black" and "bold black" are usually rendered with different RGB values (#000000 for the former, some shade of grey for the latter).  That's what xterm(1) does by default, for example.  What terminal emulator do you use?  What does `print -P %F{black}%Blorem ipsum%f%b` look like in it?  And `echo $'\e[90mlorem ipsum\e[0m'`? (I don't have handy the symbolic name for this color, if any)

Silejonu:
> any other application may use "bold black" as well. 

But from my experience, none has ever done that in practice, for good reasons.

I use GNOME Console and GNOME Terminal. The issue is present on both.

Here is the output of the commands your wrote:  
- in GNOME Console
Capture d’écran du 2023-03-31 09-55-44
Capture d’écran du 2023-03-31 10-02-39
- and in GNOME Terminal
Capture d’écran du 2023-03-31 09-58-55
Capture d’écran du 2023-03-31 10-02-18

phy1729:
I'm reluctant to change the default as it's been this way for nearly 8 years. Changing it now (especially from a relatively muted bold black to a rather loud magenta background) would be a breaking change for anyone who relies on the default. The default won't be able to please everyone, but that's why the styles are configurable :slightly_smiling_face:.

Issue #930, opened 2023-06-13
main highlighter does not highlight hashed directories with autocd and cdablevars set


zaidhaan:
Screenshot:
![image](https://github.com/zsh-users/zsh-syntax-highlighting/assets/5954994/a62f8be8-7335-4a44-8cd6-c17ee70a6e0d)

I was able to fix it with this
```diff
diff --git a/highlighters/main/main-highlighter.zsh b/highlighters/main/main-highlighter.zsh
index f691051..505fe3c 100644
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -196,6 +196,8 @@ _zsh_highlight_main__type() {
       REPLY=builtin
     elif (( $+commands[(e)$1] )); then
       REPLY=command
+    elif (( $+nameddirs[(e)$1] )) && [[ -x $nameddirs[$1] ]]; then
+      REPLY=hashed
     # None of the special hashes had a match, so fall back to 'type -w', for
     # forward compatibility with future versions of zsh that may add new command
     # types.

```

And get these results:
![image](https://github.com/zsh-users/zsh-syntax-highlighting/assets/5954994/17f5def3-5820-44ae-b42e-4b6aa63062f0)

Though I'm not sure if that's the right approach, or whether that addresses it in the right place.

**Additionally,** `cdablevars` appears to also not be respected (not sure whether this should be moved to a new issue but these both seem related so I'll keep it here)
![image](https://github.com/zsh-users/zsh-syntax-highlighting/assets/5954994/debe6368-75c1-40a8-a6c0-0ad4ae585c23)

Even with my patch above, it only is able to _partially_ addresses this, since it appears that after entering into that directory, only then does it update the hash table (and once the hash table is updated, it gets highlighted, as shown below)
![image](https://github.com/zsh-users/zsh-syntax-highlighting/assets/5954994/b6d8b8bb-05c9-4060-b9bb-9e7cee1e6a51)

---

- zsh 5.9 (x86_64-pc-linux-gnu)
- zsh-syntax-highlighting commit 1386f12

phy1729:
I think the right place to add the check would be in `_zsh_highlight_main_highlighter_check_path` where autocd is handled currently. To check the user's `cdablevars` setting you can use the `zsyh_user_options` array as done at the top of that function for `autocd`.

zaidhaan:
Ah, thanks. Looks like my fix can do the following only:
[...]

PR #931, opened 2023-06-14
main: Add highlighting for hashed directories with autocd and cdablevars

zaidhaan:
See #930 (or at least just https://github.com/zsh-users/zsh-syntax-highlighting/issues/930#issuecomment-1590633859).

What this PR does:
* If a hashed directory `foo` exists and the path it resolves to also exists, it will highlight it as a `hashed-command`
image

What this PR does not do:
* Does not highlight after further Tab-complete expansion of the hashed directory, even though it should be valid
image
image

If this feels like enough to be a PR on its own, then you could merge. If it feels incomplete, but on the right path, someone else could offer guidance and I'll see if I can complete it or they could pick it up themselves, or if it's completely wrong then it could just be closed.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -1215,6 +1215,12 @@ _zsh_highlight_main_highlighter_check_path()
     integer autocd=0
   fi
 
+  if [[ $zsyh_user_options[cdablevars] == on ]]; then
+    integer cdablevars=1
+  else
+    integer cdablevars=0
+  fi
+
   if (( in_command_position )); then
     # ### Currently, this value is never returned: either it's overwritten
     # ### below, or the return code is non-zero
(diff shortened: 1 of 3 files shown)


muratkorkmaz106:
Starting

phy1729:
I think we can take the CDABLE_VARS documentation fairly literally

              If the argument to a cd command (or an implied cd with the
              AUTO_CD option set) is not a directory, and does not begin with
              a slash, try to expand the expression as if it were preceded by
              a `~' (see the section `Filename Expansion').

and just call `_zsh_highlight_main_highlighter_expand_path` again, but prefixed with a tilde. Something like

```zsh
    elif (( autocd && cdablevars )) && [[ ${1[1]} != / ]]; then
      _zsh_highlight_main_highlighter_expand_path "~$1"
      if [[ -d $REPLY && -x $REPLY ]]; then
        REPLY=autodirectory
        return 0
      fi
```

That covers the case where the path continues and the case where a username is provided.

PR #945, opened 2024-02-22
Change Arch Linux repository name and add correct source path for .zshrc

alexnils:
The repository is called Extra nowadays. Updated path to conform with the arch-package and AUR package builds. Also updated the URL for the arch-package to directly go to the package page instead of a search.


mammadu (approved):
I had to make the same change myself. LGTM

PR #948, opened 2024-03-13
docs: Remove Fig instructions

z0rc:
This project is dead. https://fig.io/blog/post/fig-is-sunsetting

Effectively it's a revert of https://github.com/zsh-users/zsh-syntax-highlighting/commit/b2c910a85ed84cb7e5108e7cb3406a2e825a858f.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -107,12 +107,6 @@ This list is incomplete as there are too many
 Add `antigen bundle zsh-users/zsh-syntax-highlighting` as the last bundle in
 your `.zshrc`.
 
-#### [Fig](https://fig.io)
-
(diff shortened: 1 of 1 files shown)


brendanfalk:
To clarify, project isn't "dead", it's being merged into [CodeWhisperer for command line](https://docs.aws.amazon.com/codewhisperer/latest/userguide/command-line-getting-started-installing.html). However, yes, the plugins feature is not being merged into CodeWhisperer. We were planning on doing a review of all the readmes  where Fig was listed removing it, however, not for about 3 months. Fig is being sunset on 1 september and will be fully supported until then.

Nonetheless, your call on when you want to remove this

PR #949, opened 2024-03-22
Docs: Remove Bloat in the Plain Install

redkittty:
Hello,
The way you install the plugin on a plain .zshrc is quite simple, these are just some changes to debloat the .zshrc

The first change is to replace ./zsh-syntax-highlighting/zsh-syntax-highlighting.zsh with ~/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh

This is due to the fact that with ./ , zsh was looking inside the current folder which would often introduce bloat as the folders outside of home were stored and if I were to go into a different folder and open a new terminal tab, well then that folder all of the sudden has the plugin folder wasting disk space.

With ~/ , it fixes that issue because it tells zsh to look in the home folder and then to go into the plugin directory, this has less bloat because if I am in another folder, it doesn't make a folder for the plugin there as now it doesn't have to as zsh now knows it is in the home directory.

The other change is to the echo command after git clone, I put optional there because if you plain install it, you don't need that command and it ends up being bloat. The reason I added the Optional tag is because it might be needed for a specific plugin manager even though the section is titled: In your .zshrc.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -71,13 +71,17 @@ Simply clone this repository and source the script:
 
 ```zsh
 git clone https://github.com/zsh-users/zsh-syntax-highlighting.git
+```
+
+Optional:
+```zsh
 echo "source ${(q-)PWD}/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh" >> ${ZDOTDIR:-$HOME}/.zshrc
 ```
 
   Then, enable syntax highlighting in the current interactive shell:
 
 ```zsh
-source ./zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
+source ~/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
 ```
 
   If `git` is not installed, download and extract a snapshot of the latest


danielshahaf:
You seem to have copied the `git clone` and `echo` lines into your zshrc.  That's not what you were supposed to do; you were supposed to run them in an interactive shell.  Would you like to re-do the PR to clarify this point?  The PR as it stands is plain wrong, for starters because ~/zsh-syntax-highlighting might not exist.

The incumbent `source` line is correct too.  It, too, is meant to be pasted /into a running shell/, not into any file.

Thanks for the report.

Excuse brevity.

redkittty:
oh I see, I just did this because I often git clone in the home directory, so if I went to another directory, It would make a directory inside that other directory I was working in. I was just confused because in bold text it said, In your ~.zshrc. You could maybe bold the text for doing the source inside the current shell

danielshahaf:
It's in bold because it's a section header.  I can see how it's confusingm though.  If someone (not necessarily you) has a minute to change the text to prevent this misunderstanding, that would be a good thing.

PR #958, opened 2024-09-23
main: Add systemd run0 equivalent to sudo

ColinKinloch:
~I suspect these changes are incorrect or incomplete.
Is the second value a list of single letter flags arguments? Should it be `Cgprtu:AEHPSbilns:eKkVv`?~
[`run0`](https://www.freedesktop.org/software/systemd/man/devel/run0.html) is a systemd 256 alternative to `sudo`.

### DIFF
--- a/highlighters/main/main-highlighter.zsh
+++ b/highlighters/main/main-highlighter.zsh
@@ -354,6 +354,7 @@ _zsh_highlight_highlighter_main_paint()
     'pkexec' '' # doesn't take short options; immune to #121 because it's usually not passed --option flags
     # Not listed: -h, which has two different meanings.
     'sudo' Cgprtu:AEHPSbilns:eKkVv # as of sudo 1.8.21p2
+    'run0' ugD:h # systemd 256.6
     'stdbuf' ioe:
     'eatmydata' ''
     'catchsegv' ''


danielshahaf:
> Is the second value a list of single letter flags arguments?

The middle value is list of single-letter flags that don't take arguments.  E.g., 'sudo -u foo' uses 'foo' as an argument to '-u', so sudo's "u" is before the first colon; but 'sudo -l foo' interprets 'foo' as a positional argument, hence sudo's '-l' is in the middle group.

Lines 35 through 337 were supposed to make this clear, but there's room for improvement.

ColinKinloch:
That makes sense, thanks.
In that case the only options are `ugD` each taking an argument.
And `-h` which seems to be intentionally omitted.

danielshahaf:
> And `-h` which seems to be intentionally omitted.

I do see a -h in <https://man.archlinux.org/man/run0.1.en>.

Could you add a comment stating the run0 version referenced?  That's helpful down the road when looking for what flags were added/changed (we can just go through the upstream changelog, or diff upstream's man page source between tags).

Other than that, LGTM, but I'll defer to others to do the actual merge.

ColinKinloch:
> I do see a -h in <https://man.archlinux.org/man/run0.1.en>.

Ah, never mind. I was confused by the absence of `h` in `sudo`. having an option switch meanings seems like a bad idea :thinking: .

I've added `h` and noted that it's `systemd 256.6`

rpigott:
Gentle ping.

I hope to add -k/-K/-v/-n to run0 in systemd/systemd#42465 fwiw, but that won't be available until v262 at the earliest. Either way this is a nice to have in z-s-h.

phy1729:
Merged. Thanks!

PR #959, opened 2024-11-01
Update INSTALL.md for Arch

ildoc:
if the package is installed through pacman, the .zsh path is `/usr/share/zsh/plugins/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh`


Scutua (approved):

reppiz:
Just used pacman to install the package and found out this true for my system too.

Issue #969, opened 2025-12-01
[Bug]: Text disappears while typing $


tausiq2003:
Description: After enabling the zsh-syntax-highlighting and add it to .zshrc and source it, in command line after typing `$` followed by any command,  the text disappears.

Steps to reproduce:
1. Install zsh-syntax-highlighting, assuming you have zsh.
2. Add it to `.zshrc`, change
```diff
-  plugins=(git)
+ plugins=(git zsh-syntax-highlighting)
```
3. source it, `source ~/.zshrc`
4. Now type anything after typing `$`, text disappears.

How it should behave: 
- Text shouldn't disappear.

Screen Recording: 
https://asciinema.org/a/To40K2TpYKoexGYEB8daq6BYt

Terminal: alacritty
zsh version: 5.9
OS: NixOS and Docker(ubuntu)
Output of `echo "$OSTYPE | $VENDOR | $MACHTYPE | $CPUTYPE | $(uname -m -p)"`: linux-gnu | pc | x86_64 | x86_64 | x86_64 x86_64

danielshahaf:
The text doesn't "disappear"; it's still there, but in black / dark grey.

By default that colour is only used for comments (when INTERACTIVE_COMMENTS is set).  So, you might want to double check that the standard 8 colours render as something readable, both bold and non-bold variants (so 16 colour variants).

Also, by default, `$echo` (with dollar) in command position would be highlighted according to the value of the variable literally named "echo".  I suspect it's not z-sy-h doing that colouring.  Does the problem occur in `zsh -f` after you run `source /path/to/zsh-syntax-highlighting.zsh` at the shell prompt?

Thanks for all the details.

tausiq2003:
Thanks for replying, its not grey its black, as in the source code I checked, but many people use black or dark mode terminal. So, I guess there is no point to make it black, also this is for comments, both (INTERACTIVE_COMMENTS (`#`)  and (`$`) or "elided parameters in command position" as per the documentation, and as to answer your question does this happen in `zsh -f`, no it doesn't, so as I understood both `#` and `$` acts as comments and ZSH_HIGHLIGHT_STYLES has "fg=black,bold" and a user can set it manually, but, it would be good if it will be more like "fg=240,bold", let me know what you think

danielshahaf:
[...]

PR #970, opened 2025-12-08
fix: changed color of ZSH_HIGHLIGHT_COLOR[comment]

tausiq2003:
So, if we type comments (interactive comments (`#`) or "elided parameters in command position" (`$`)), then its fg is black which is not visible for the normal terminal users, so it doesn't makes any sense, so I changed it a greyish version to 245 so that it will be clearly visible, fixed issue #969


highlighters/main/main-highlighter.zsh
```diff
@@ -59,7 +59,7 @@
 : ${ZSH_HIGHLIGHT_STYLES[back-dollar-quoted-argument]:=fg=cyan}
 : ${ZSH_HIGHLIGHT_STYLES[assign]:=none}
 : ${ZSH_HIGHLIGHT_STYLES[redirection]:=fg=yellow}
-: ${ZSH_HIGHLIGHT_STYLES[comment]:=fg=black,bold}
+: ${ZSH_HIGHLIGHT_STYLES[comment]:=fg=245,bold}
```
danielshahaf, line 62: This assumes the terminal supports 256 colours.  Is this assumption valid in all supported use-cases?

(It also assumes _zsh_ supports that syntax, but I don't have time right now to look up when the `fg=`_N_ syntax was added to zsh.  Probably long enough ago.)

danielshahaf:
To the other maintainers: I'm not sure whether I should be pressing "Approve workflows to run" right now or not.  If workflows should be approved to run, could someone greenlight them, please?

highlighters/main/main-highlighter.zsh
tausiq2003: didn't tested it locally, but I changed the `ZSH_HIGHLIGHT_STYLES[comment] to fg=245,bold and it worked in my system(nixos), and ubuntu on docker, zsh version: 5.9

PR #980, opened 2026-06-06
Highlight Git subcommands

Iamliuxiaozhen:
## Summary

This PR adds minimal semantic highlighting for Git subcommands in the main highlighter.

Currently, `git` itself is highlighted as a valid command, but the following subcommand is not checked:

```zsh
git status
git statu
```
With this change:

known Git subcommands are highlighted as command
unknown Git subcommands are highlighted as unknown-token
the Git subcommand list is loaded lazily and cached
external git-* commands are also considered
Scope

This is intentionally limited to the simple form:

git <subcommand>

It does not yet handle:

git -C repo status
git --git-dir=.git status
git -c key=value status

It also does not validate subcommand arguments or options.

Tests

Adds coverage for valid and invalid Git subcommands.


phy1729:
While `git` is a common command with subcommands, adding a helper function and this/next_word state per command with subcommands will get unruly quickly. Could you look at making this more generic so it could handle some predefined and possibly even user defined commands?

Also this doesn't handle the case of command flags before the subcommand (consider `git --git-dir ~/.dotfiles --work-tree . status`).

PR #986, opened 2026-08-21
tests: fix relative links in tests/README.md

yfwmaniish:
## Summary
- `tests/README.md` links to `tests/test-highlighting.zsh` and `tests/test-perfs.zsh`, but this file already lives in `tests/`, so those links resolve to `tests/tests/test-highlighting.zsh` and `tests/tests/test-perfs.zsh` — both 404 on GitHub right now.
- Drops the redundant `tests/` prefix so both links resolve relative to the file's actual location.

## Test plan
- [x] Confirmed with a small link-checker script (extracts markdown inline links, resolves them relative to the containing file, flags any that don't exist on disk) — it flagged both links before the fix and reports clean after.


phy1729:
Thanks!

PR #989, opened 2026-08-28
ci: point apt at archive.debian.org so the test matrix can install

yfwmaniish:
Every job in the test matrix has been failing since Debian 10 "buster" was archived — on `master` as well as on pull requests. This restores it.

## What's happening

The matrix runs inside `ghcr.io/zsh-users/zsh:${{ matrix.version }}`. All 35 of those images are buster-based — `/etc/debian_version` reads `10.12`, and `/etc/apt/sources.list` is:

```
deb http://deb.debian.org/debian buster main
deb http://security.debian.org/ buster/updates main
```

buster has since been dropped from both of those mirrors:

| URL | Status |
| --- | --- |
| `deb.debian.org/debian/dists/buster/Release` | **404** |
| `archive.debian.org/debian/dists/buster/Release` | 200 |
| `archive.debian.org/debian-security/dists/buster/updates/Release` | 200 |

`install_packages` is minideb's `apt-get update && apt-get install` helper, which retries three times and then exits with apt's status. So every job dies like this, before `make test` is ever reached:

```
E: The repository 'http://deb.debian.org/debian buster Release' does not have a Release file.
E: The repository 'http://security.debian.org buster/updates Release' does not have a Release file.
apt failed, retrying
...
##[error]Process completed with exit code 100.
```

Nothing in the test suite is involved, which is why the failure is uniform from 4.3.11 through master.

## The fix

Repoint the two entries at `archive.debian.org`, which still serves buster — including `buster/updates` under `debian-security` — and still carries `bsdmainutils`, `make` and `procps`, verified against the buster `main` package index.

The rewrite is guarded on `sources.list` actually mentioning buster, so it turns into a no-op the moment the images are rebuilt on a supported release, rather than silently redirecting a live suite to the archive.

## Notes

- No `Acquire::Check-Valid-Until` override is needed here: the archived buster `Release` carries a `Date` but no `Valid-Until` field, so apt won't treat it as expired.
- Rebuilding the container images on a supported Debian would be the better long-term fix. This is the in-repo change that unblocks CI in the meantime, and it retires itself when that happens.


phy1729:
This should be fixed by updating the base image not by using the archive.

